PR #54, opened 2024-07-13
Add no-segments-on-sliced-layers rule

### LINKED ISSUE #25: Implement `no-segments-on-sliced-layers`
Sliced layers (Entities, Features, Widgets, Pages) should not have folders with conventional segment names in direct children.

- [ ] Write the rule
- [ ] Write positive test
- [ ] Write negative test
- [ ] Add documentation in form of a README and link that in the main README of the repo

daniilsapa:
Resolves #25


illright (changes requested):
Hi, thanks for the PR! I looked over it quickly, thank you for including the changeset and updates to the README :) 

I ran the tests and they are failing on Windows. It might have to do with the fact that you split by `'/'`, because Windows paths are separated with `\`. You can use `sep` from `node:path` to fix that

daniilsapa:
Yes, this is most likely the cause of the problem. Thank you. Fixing the PR

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/README.md
```diff
@@ -0,0 +1,47 @@
...
+    📄 index.ts
+📂 pages
+   📂 home
+     📂 ui
+     📄 index.ts
+```
+
+## Rationale
+
+Slices exist to partition code by business domain and entities. You can freely create and name them (e.g. `pages` home, profile and `entities` user, product, ...) based on your needs, application logic, company glossary, etc. Slices contain code of different type/purposes (segments) to implement their part of functionality. Segments (`ui`, `lib`, `api`) are simply a division of code by purpose, thus "ownerless" segments in sliced layers are not allowed since they need to be attached to some part of the business domain inside these layers.
```
illright, line 47: suggestion: I would extend this rationale a bit. You are describing the case when a folder `entities/api` was created as a segment, but there can also be cases when `entities/api` was meant to be a slice. Both cases are confusing, so we want to discourage them both. Here's my version of the rationale that describes both of these issues:
```suggestion
Several folder names like `ui` or `api` are conventionally understood in FSD as segments. When you have segments as direct children, it's either because that's the name you chose for your slice, or because some code ended up unsliced. In the first case, where that is the name of your slice, such a name is likely to cause confusion among developers who are familiar with FSD. In the second case, where a segment on a layer is because there was no slice to put it in, it's a violation of FSD structure, which decreases the benefits of FSD and also causes confusion.
```

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: 'no-segments-on-sliced-layers',
+  check(root) {
+    const diagnostics: Array<Diagnostic> = []
+    const layers = Object.values(getLayers(root))
+
+    for (const layer of layers) {
+      if (isSliced(layer)) {
+        for (const directChild of layer.children) {
+          if (directChild.type === 'folder') {
+            const folderName = directChild.path.split(sep).pop()
```
illright, line 16: suggestion (non-blocking): you can use `basename` from `node:path` to get the folder name:
```suggestion
            const folderName = basename(directChild.path)
```

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/index.ts
```diff
@@ -0,0 +1,34 @@
...
+            const isConventionalSegment = folderName && conventionalSegmentNames.includes(folderName)
+
+            if (isConventionalSegment) {
+              diagnostics.push({
+                message: `Conventional segment "${folderName}" found as a direct child of a sliced layer. Consider moving it inside a slice or to another layer.`,
```
illright, line 21: issue: I'm not sure what is meant by "consider moving it to another layer"

suggestion: as I described in a previous comment, there can be two cases — when the intention was for that folder to be a slice, or when the intention was for it to be a segment. If it was supposed to be a slice, a good suggestion, in my opinion, would be to rename it. If it was meant to be a segment, then the suggestion should be to move it inside a slice.
```suggestion
                message: `Conventional segment "${folderName}" should not be a direct child of a sliced layer. Consider moving it inside a slice or, if that is a slice, consider a different name for it to avoid confusion with segments.`,
```

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/index.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+          📄 index.ts
+        📂 settings
+          📂 profile
+            📄 ProfilePage.tsx
+            📄 index.ts
+        📂 lib
+          📄 index.ts
+    `)
+
+    const diagnostics = noSegmentsOnSlicedLayers.check(root).diagnostics
```
illright, line 74: suggestion (non-blocking): it would be better for the test to sort these diagnostics so that sort order doesn't break the test under a different implementation. We have a helper called `compareMessages` that you can pass to `.sort()` as a comparator function

illright (changes requested):
Fantastic job! I'm very happy that you did all the steps that are needed to add a rule. I also like the code of the rule, it's short and to the point.

I've left some suggestions, mostly around the error messages, as that is the most important part of a rule. Let me know what you think about them

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/README.md
daniilsapa: Yes, I didn't foresee that the user could choose api as the name of a slice

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/index.ts
daniilsapa: 👍
Yes, it looks more concise, graceful and does not reinvent the wheel

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/index.ts
daniilsapa: Yes, my "consider moving it to another layer" sounds a bit misleading because what I meant was more "consider moving it to the shared layer" for the case where the user puts some highly reusable code that is not related to any slice

packages/steiger-plugin-fsd/src/no-segments-on-sliced-layers/index.spec.ts
daniilsapa: Yes, thank you for the suggestion

daniilsapa:
Thanks for your feedback! I'm glad to bring value to the project.  I agree, your suggestions make perfect sense and improve the solution. I have already corrected the highlighted parts

illright (approved):

illright:
I guess `toSorted` is not available in Node 18. Could you replace it with `.sort()`?

daniilsapa:
Yep, sure

daniilsapa:
@illright 
Done

PR #57, opened 2024-07-16
Add app layer to exceptions of no-layer-public-api rule

### LINKED ISSUE #55: Add an exception for an index file inside the App layer
Right now it reports an error on such a case. But it looks like it should not, because the app layer is the entry point to your application, and it's pretty intuitive that it could have an index.js/ts file.

Screenshot 2024-07-13 at 23 35 16

daniilsapa:
Resolves #55

### DIFF
--- a/packages/steiger-plugin-fsd/src/no-layer-public-api/README.md
+++ b/packages/steiger-plugin-fsd/src/no-layer-public-api/README.md
@@ -11,6 +11,8 @@ According to the _public API rule on slices_:
 
 A corollary to this rule is that the layer itself should not have an index file.
 
+**Exception:** index files are allowed on `app` layer because it's the entry point of the application.
+
 Examples of project structures that pass this rule:
 
 ```
--- a/packages/steiger-plugin-fsd/src/no-layer-public-api/README.md
+++ b/packages/steiger-plugin-fsd/src/no-layer-public-api/README.md
@@ -11,7 +11,7 @@ According to the _public API rule on slices_:
 
 A corollary to this rule is that the layer itself should not have an index file.
 
-**Exception:** index files are allowed on `app` layer because it's the entry point of the application.
+**Exception:** index files are allowed on `app` layer because some people prefer to have that as their app's entrypoint.
 
 Examples of project structures that pass this rule:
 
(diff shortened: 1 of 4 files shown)


daniilsapa:
@illright 
Could you tell me how I can test Steiger locally? How do you usually do that? I tried to run `turbo build` and then run dist/cli.js file but got an error.

illright:
Weird, what error did you get?

packages/steiger-plugin-fsd/src/no-layer-public-api/README.md
```diff
@@ -11,6 +11,8 @@ According to the _public API rule on slices_:
 
 A corollary to this rule is that the layer itself should not have an index file.
 
+**Exception:** index files are allowed on `app` layer because it's the entry point of the application.
```
illright, line 14: I would rephrase this to say "because some people prefer to have that as their app's entrypoint" to emphasize that this is not a rule, but a preference

.changeset/fair-ducks-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@feature-sliced/steiger-plugin': minor
```
illright, line 2: I think this qualifies more as a patch change, since it's not adding new features

packages/steiger-plugin-fsd/src/no-layer-public-api/index.ts
```diff
@@ -1,6 +1,9 @@
 import { getIndex, getLayers } from '@feature-sliced/filesystem'
 import type { Diagnostic, Rule } from '@steiger/types'
 
+/** Layers that are allowed to have an index file. */
```
illright, line 4: praise: thanks for appropriately documenting your code!

illright (changes requested):
Another great PR, thanks! A few corrections from my side

.changeset/fair-ducks-hope.md
daniilsapa: Indeed, you’re right

daniilsapa:
The error occurs when you try to specify a folder that is outside the working folder. This is true for both local and distributed app builds.

Here's what it looks like

Screenshot 2024-07-17 at 16 23 09

illright (approved):
Thanks for implementing the corrections! As for the issues you're facing in development, it's weird, I've never seen it before. Could you make an issue about it, so that we don't lose this conversation?

daniilsapa:
You're welcome. Yes, I created an issue for that error

PR #59, opened 2024-07-17
Exit with code 1 when errors are found

### LINKED ISSUE #56: Exit with 1 exit code when there are reported errors
I tried to integrate Steiger into a pre-commit validation workflow and I failed. The reason is that it always exits with 0 code even if there are some errors. 0 means success, so it does not fail the pre-commit check when errors are found.

daniilsapa:
Resolves #56

### DIFF
--- a/.changeset/three-crews-end.md
+++ b/.changeset/three-crews-end.md
@@ -1,0 +1,5 @@
+---
+'steiger': patch
+---
+
+Make Steiger exit with 1 if errors are reported
--- a/packages/steiger/src/cli.ts
+++ b/packages/steiger/src/cli.ts
@@ -74,9 +74,15 @@ if (consoleArgs.watch) {
   })
 } else {
   const diagnostics = await linter.run(resolve(consoleArgs._[0]))
+  let stillRelevantDiagnostics = diagnostics
 
   reportPretty(diagnostics, process.cwd())
+
   if (consoleArgs.fix) {
-    applyAutofixes(diagnostics)
+    stillRelevantDiagnostics = await applyAutofixes(diagnostics)
+  }
+
+  if (stillRelevantDiagnostics.length > 0) {
+    process.exit(1)
   }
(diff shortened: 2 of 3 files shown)


daniilsapa:
Hi @illright 
Could you review it when you have time?

I had to change the autofixing logic a bit to accomplish this task.
Also I noticed several issues with autofix engine and also possible improvements

Issues that I noticed so far:
- It autofixes with .js files even if a project uses TS
- Why do we print errors first and then try to autofix them, shouldn't we swap these 2 actions to not report errors that were autofixed?

I saw open tickets to improve the autofix mechanism and realized that this was a separate complex problem within this project.

Possible issues/improvements that I think are worth discussion:
- What if a file was moved with one fix and then we run another fix that needs to fix imports inside that file? As I see right now the autofix engine does not handle such a case
- It's better to have some kind of abortion logic, for example, to handle cases when a complex fix is applied partially and then fails

I think we need to manipulate some kind of virtual file system and then when we are sure that a needed fix (sequence of fixes, e.g. "move" -> "rename" -> "fix import statements inside") is possible to apply without errors, we commit it to the real FS. 

Let's create a separate discussion. Because it feels like a complex problem that needs a lot of discussions

illright (changes requested):
Looks good, thank you! As for the auto-fix engine, yes, it's a whole other story. I have some ideas in mind about what to do with it, but let's talk about it in one of the relevant issues.

Could you also add a patch changeset for Steiger please?

daniilsapa:
Oh, yes, sorry. I forgot to push that commit yesterday

daniilsapa:
@illright 
I created a separate discussion where I drafted my vision/ideas of the engine, so we can groom it in one place and not scatter around the auto-fix-related issues. Looking forward to seeing your ideas on the auto-fix engine there!

illright (approved):

PR #64, opened 2024-07-21
Make sure forbidden-imports rule checks files directly inside layers

### LINKED ISSUE #49: Bug: there is no error about forbidden vertical imports
May be I'm doing something wrong, but I can't get an error about forbidden import hierarchy, e.g. when importing `app` in `entities`.
Here is the sandbox: https://codesandbox.io/p/devbox/steiger-test-g4f4wp

daniilsapa:
Resolves #49 

Had to move all layers into the /src folder for the mock filesystem used in the unit tests for the rule because there was an edge case where ts.resolveModuleName could not correctly resolve aliases for folders that are directly at the root of the filesystem (it did not resolve /app/index.ts from @/app and same problem for all layers). After moving everything in /src, they resolve just fine. I thought about creating an issue in TS repository, but it seems like an extremely rare edge case that nobody would care about. No one puts their real projects in the root of the file system.

### DIFF
--- a/packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
+++ b/packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
@@ -31,7 +31,9 @@ export function indexSourceFiles(root: Folder): Record<string, SourceFile> {
     // Even though files that are directly inside a layer are not encouraged by the FSD and are forbidden in most cases
     // (except for an index/root file for the app layer as an entry point to the application), users can still add them.
     // So, we need to index all files directly inside a layer to find errors.
-    walk(layer, { layerName: layerName as LayerName, sliceName: null, segmentName: null })
+    layer.children
+      .filter((child) => child.type === 'file')
+      .forEach((file) => walk(file, { layerName: layerName as LayerName, sliceName: null, segmentName: null }))
 
     if (!isSliced(layer)) {
       for (const [segmentName, segment] of Object.entries(getSegments(layer))) {
--- a/packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
+++ b/packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
@@ -77,6 +77,11 @@ if (import.meta.vitest) {
             📄 EditorPage.tsx
             📄 Editor.tsx
           📄 index.ts
+      📂 app
+        📂 ui
+          📄 index.ts
+        📄 root.ts
+        📄 index.ts
     `)
 
     expect(indexSourceFiles(root)).toEqual({
@@ -168,6 +173,33 @@ if (import.meta.vitest) {
         },
         layerName: 'shared',
         segmentName: 'ui',
+        sliceName: null,
+      },
+      [joinFromRoot('app', 'ui', 'index.ts')]: {
+        file: {
+          path: joinFromRoot('app', 'ui', 'index.ts'),
+          type: 'file',
+        },
+        layerName: 'app',
+        segmentName: 'ui',
+        sliceName: null,
+      },
+      [joinFromRoot('app', 'root.ts')]: {
+        file: {
+          path: joinFromRoot('app', 'root.ts'),
+          type: 'file',
+        },
+        layerName: 'app',
+        segmentName: 'root',
+        sliceName: null,
+      },
+      [joinFromRoot('app', 'index.ts')]: {
+        file: {
+          path: joinFromRoot('app', 'index.ts'),
+          type: 'file',
+        },
+        layerName: 'app',
+        segmentName: null,
         sliceName: null,
       },
     })
--- a/packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
+++ b/packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
@@ -28,6 +28,11 @@ export function indexSourceFiles(root: Folder): Record<string, SourceFile> {
   }
 
   for (const [layerName, layer] of Object.entries(getLayers(root))) {
+    // Even though files that are directly inside a layer are not encouraged by the FSD and are forbidden in most cases
+    // (except for an index/root file for the app layer as an entry point to the application), users can still add them.
+    // So, we need to index all files directly inside a layer to find errors.
+    walk(layer, { layerName: layerName as LayerName, sliceName: null, segmentName: null })
+
     if (!isSliced(layer)) {
       for (const [segmentName, segment] of Object.entries(getSegments(layer))) {
         walk(segment, { layerName: layerName as LayerName, sliceName: null, segmentName })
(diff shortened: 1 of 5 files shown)


packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
```diff
@@ -28,6 +28,11 @@ export function indexSourceFiles(root: Folder): Record<string, SourceFile> {
   }
 
   for (const [layerName, layer] of Object.entries(getLayers(root))) {
+    // Even though files that are directly inside a layer are not encouraged by the FSD and are forbidden in most cases
+    // (except for an index/root file for the app layer as an entry point to the application), users can still add them.
+    // So, we need to index all files directly inside a layer to find errors.
+    walk(layer, { layerName: layerName as LayerName, sliceName: null, segmentName: null })
```
daniilsapa, line 34: @illright 
Actually, you can postpone your review. I will fix the problem with this line. Because I just realized that it will go through all nested files/directories. And then we walk through them again in the lines below

daniilsapa:
@illright 
I fixed the problem that I introduced. But still have a question. Based on the logic of `indexSourceFiles`, currently, we index files based on the ideal folder structure according to FSD. But in the real world users sometimes can put files in unexpected (incorrect in terms of FSD) places. The question is, should we still index such files to find errors inside or report the placement of the file itself as an error?

packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
```diff
@@ -28,6 +28,13 @@ export function indexSourceFiles(root: Folder): Record<string, SourceFile> {
   }
 
   for (const [layerName, layer] of Object.entries(getLayers(root))) {
+    // Even though files that are directly inside a layer are not encouraged by the FSD and are forbidden in most cases
+    // (except for an index/root file for the app layer as an entry point to the application), users can still add them.
+    // So, we need to index all files directly inside a layer to find errors.
+    layer.children
+      .filter((child) => child.type === 'file')
+      .forEach((file) => walk(file, { layerName: layerName as LayerName, sliceName: null, segmentName: null }))
```
illright, line 36: suggestion: could you add a test for this case please? The tests are in the same file, in the `if (import.meta.vitest)` thing

packages/steiger-plugin-fsd/src/forbidden-imports/index.ts
```diff
@@ -1,12 +1,13 @@
 import * as fs from 'node:fs'
 import { layerSequence, resolveImport } from '@feature-sliced/filesystem'
 import precinct from 'precinct'
-const { paperwork } = precinct
```
illright, line 4: issue: this was deliberately written like that. It's an unfortunate limitation of esbuild that we have to work around by doing this, and the point is to show that the import is actually called `paperwork`, it's just that esbuild cannot recognize it

suggestion: roll this change back

packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
```diff
@@ -16,103 +28,158 @@ vi.mock('node:fs', async (importOriginal) => {
...
   const root = parseIntoFsdRoot(`
-    📂 shared
-      📂 ui
-        📄 styles.ts
-        📄 Button.tsx
-        📄 TextField.tsx
-        📄 index.ts
-    📂 pages
-      📂 editor
+    📂 src
```
illright, line 60: issue: the idea is to give the `check` function of rules the folder of the FSD root, so by putting the layers in `src`, we're not actually testing how it's going to be

suggestion: we can add a second optional argument to `parseIntoFsdRoot` — the root path of the folders, then we can avoid having `src` here in the structure

illright: ah, now I see that in other tests, you access `.children[0]` to overcome this issue. That's also fine, although you forgot to do it in the positive case. I still think that it would be cleaner with an optional root path argument though

illright (changes requested):

packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
daniilsapa: Yeah, I had to put all the files into /src because of the problem I wrote about in the PR description. Indeed I forgot to add `.children[0]` for the positive case, thanks for catching that. 

I would be happy to pass that optional argument to `parseIntoFsdRoot` but I see that it accepts only 1 argument. Maybe you implemented the second parameter in some branch that has not yet been merged to the master?

Screenshot 2024-07-23 at 22 36 08

packages/steiger-plugin-fsd/src/_lib/index-source-files.ts
daniilsapa: Yes, done

packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
illright: I meant to suggest for you to add that second parameter :)

illright: I also took the liberty to unresolve this conversation. Usually, I prefer to mark threads as resolved when the comment was either fully addressed or it was decided not to address it, but this comment is still relevant partially. Let me know if you'd like to add that second parameter, or I can do it

daniilsapa: Oh, got it. No, problem, I'll add that tomorrow. I think it will be useful in further test cases and also I don't like `.children[0]` too much.

packages/steiger-plugin-fsd/src/_lib/prepare-test.ts
```diff
@@ -4,8 +4,31 @@ import type { FsdRoot } from '@feature-sliced/filesystem'
...
+          }
+        }
+      }
+    }
+
+    throw new Error(`Path "${path}" not found in the provided file system mock!`)
+  }
+
+  return checkFolder(folder)
+}
```
illright, line 28: suggestion: I meant a more simple solution — you pass the root path as the second argument, and whatever file tree you write with emojis is "mounted" to that root. For example:

```ts
parseIntoFsdRoot(`
  📂 test
    📄 file
`, "/home/illright");  // the "file" becomes `/home/illright/test/file`
```

illright (changes requested):

packages/steiger-plugin-fsd/src/_lib/prepare-test.ts
daniilsapa: Yeah, I agree, the simpler the better, the [KISS](https://www.youtube.com/watch?v=ZhIsAZO5gl0) principle should not be forgotten 😂

illright (approved):
Okay, we're ready to merge this. Thanks a lot for doing all the corrections!

daniilsapa:
No problem, thanks for your reviews and feedback!

PR #74, opened 2024-07-30
 Make the rule set configurable with plugins

### LINKED ISSUE #67: Make the rule set configurable with plugins
Currently, the default ruleset is hard-coded. Refactor the FSD plugin to instead export a plugin (see https://github.com/feature-sliced/steiger/discussions/53#discussioncomment-9980110 for the type definition) and make Steiger register rules from plugins.

Make sure to alert about collisions in rule names — they must be unique. To that end, rename all existing rules in the FSD plugin to include an `fsd/` prefix in the rule name.

The default config should include the FSD plugin and enable all of its rules. Create `configs.recommended` with all rules enabled

daniilsapa:
Resolves #67

### DIFF
--- a/packages/steiger-plugin-fsd/src/index.ts
+++ b/packages/steiger-plugin-fsd/src/index.ts
@@ -14,7 +14,10 @@ import segmentsByPurpose from './segments-by-purpose/index.js'
 import sharedLibGrouping from './shared-lib-grouping/index.js'
 import noProcesses from './no-processes/index.js'
 
-export default [
+import packageJson from '../package.json'
+import { Config, Rule, Plugin, ConfigObject } from '@steiger/types'
+
+const allRules: Array<Rule> = [
   ambiguousSliceNames,
   excessiveSlicing,
   forbiddenImports,
@@ -31,3 +34,19 @@ export default [
   sharedLibGrouping,
   noProcesses,
 ]
+
+const allRulesEnabledConfig: ConfigObject = {
+  rules: allRules.reduce((acc, rule) => ({ ...acc, [rule.name]: 'error' }), {}),
+}
+
+export const plugin: Plugin = {
+  meta: {
+    name: 'steiger-plugin-fsd',
+    version: packageJson.version,
+  },
+  ruleDefinitions: allRules,
+}
+
+export const configs: Record<string, Config> = {
+  recommended: [plugin, allRulesEnabledConfig],
+}
--- a/packages/steiger-plugin-fsd/src/index.ts
+++ b/packages/steiger-plugin-fsd/src/index.ts
@@ -1,3 +1,5 @@
+import { Config, Rule, Plugin, ConfigObject } from '@steiger/types'
+
 import ambiguousSliceNames from './ambiguous-slice-names/index.js'
 import excessiveSlicing from './excessive-slicing/index.js'
 import forbiddenImports from './forbidden-imports/index.js'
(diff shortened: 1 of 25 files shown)


daniilsapa:
@illright 
I made it so that it uses the recommended default configuration from the FSD plugin if the configuration file is not found.

Also, question. Shouldn't we create a separate task to prepare a migration guide for the new config structure? Additionally, we could  
- create a script that would bring an old config to the new structure
- support the previous config with warnings that it's outdated

illright:
Regarding the migration — good point, I forgot to make a task for it. I don't think we need to support the old format, since we're not 1.0 yet, but yes, providing a best-effort migration script will be useful. I'll make a task for it.

packages/steiger-plugin-fsd/src/index.ts
```diff
@@ -31,3 +34,19 @@ export default [
...
+
+export const plugin: Plugin = {
+  meta: {
+    name: 'steiger-plugin-fsd',
+    version: packageJson.version,
+  },
+  ruleDefinitions: allRules,
+}
+
+export const configs: Record<string, Config> = {
```
illright, line 50: issue: these are named exports, and I suspect that will promote a different way of usage:
```ts
import { configs } from '@feature-sliced/steiger-plugin'

// use `configs.recommended`
```

I think this is slightly problematic because if you have two plugins, you have to do awkward renames.

suggestion: make one default export that consumers can call whatever they want. I think this code reads better:
```ts

// use `fsd.configs.recommended`
```

packages/steiger/src/models/config.ts
```diff
@@ -1,33 +1,81 @@
...
+      return { ...acc, rules: { ...acc.rules, ...item.rules } }
+    }
 
-$config.on(setConfig, (_state, payload) => payload)
+    return acc
+  }, {})
+}
+
+/**
+ * Dynamically build a validation scheme based on the rules provided by plugins.
```
illright, line 43: praise: wow, I didn't even think about this aspect, great idea!

packages/steiger/src/app.ts
```diff
@@ -1,25 +1,26 @@
...
   if (ruleConfigs === undefined) {
-    return fsdRules
+    return rules
   }
 
-  return fsdRules.filter(
+  return rules.filter(
     (rule) => !(rule.name in ruleConfigs) || ruleConfigs[rule.name as keyof typeof ruleConfigs] !== 'off',
   )
 })
```
illright, line 26: question: why do we need `$rules` here? I thought they basically contain the same things, only `$config` contains more information

illright (changes requested):
Thank you for this great PR, I'm very happy with these changes! Just one blocker on my side

packages/steiger/src/app.ts
daniilsapa: @illright 
`$rules` contains all rule definitions (registered by plugins) the tool can run. `$config` contains configurations like rule-to-severity map (in the future "files" and "ignore" globs, and other settings). So, based on that info we determine rule instances that need to be run (enabledRules)

packages/steiger-plugin-fsd/src/index.ts
daniilsapa: Yes, the current shape of exports does not correspond to that set out in the proposal document. That shape makes perfect sense. I'll make the needed changes. Thanks for catching it!

packages/steiger/src/models/config.ts
daniilsapa: 🤝 As always, I am happy to bring value to the project

illright (approved):

illright:
Great stuff!

PR #80, opened 2024-08-11
Implement inclusion/exclusion operators on VFS

### LINKED ISSUE #66: Implement inclusion/exclusion operators on a VFS
It should be a pure function that accepts a `Folder` object and a set of `include`/`exclude` globs that then produces a new `Folder` object that is a subset of the original (without modifying the original)

daniilsapa:
Resolves #66 

Hi   @illright 

Sorry for the delay. I wanted to complete this task earlier, but a terrible force majeure happened at the beginning of the week.

… but finally here’s my solution.  
Just a few annotations for it: 
- I saw that we already had a glob matching library “globby” in place but I then had to install "minimatch" because globby cannot just match file path to a pattern, it needs to traverse the real FS.
- I just implemented the functionality and didn’t connect it to other parts as we discussed it would be done in further tasks
- I had a question about test utils (and Ileft it on the specific line)

### DIFF
--- a/packages/steiger/src/features/apply-globs.test.ts
+++ b/packages/steiger/src/features/apply-globs.test.ts
@@ -1,0 +1,328 @@
+import { expect, it, describe } from 'vitest'
+
+import applyGlobs from './apply-globs'
+import { joinFromRoot, parseIntoFsdRoot } from '../../../steiger-plugin-fsd/src/_lib/prepare-test'
+
+describe('applyGlobs', () => {
+  it('should return the passed folder if no globs are provided', () => {
+    const root = parseIntoFsdRoot(
+      `
+      📂 shared
+        📂 ui
+          📄 styles.css
+          📄 Button.tsx
+          📄 TextField.tsx
+          📄 index.ts
+    `,
+      joinFromRoot('src'),
+    )
+
+    expect(applyGlobs(root, {})).toEqual(root)
+    expect(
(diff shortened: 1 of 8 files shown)


packages/steiger/src/features/apply-globs.test.ts
```diff
@@ -0,0 +1,328 @@
+import { expect, it, describe } from 'vitest'
+
+import applyGlobs from './apply-globs'
+import { joinFromRoot, parseIntoFsdRoot } from '../../../steiger-plugin-fsd/src/_lib/prepare-test'
```
daniilsapa, line 4: Should we create a shared folder for test utilities for both Steiger and the fsd-plugin? Because, maybe it's just me, but I don't really like this import

illright: I think we can just copy-paste that file, no need to set up a whole separate folder for it

illright (approved):
Hey, thanks for this, this is huge! No stress about taking more time, hope everything is alright with you.

I'm very satisfied with this PR, especially the extensive test coverage. Great job!

packages/steiger/src/features/apply-globs.test.ts
illright: Let me know if you'd like to do it before merging, because I'm honestly fine merging it as is

Issue #78, opened 2024-08-10
Widgets are falsely marked as not referenced


illright:
Source: https://github.com/AwesomeXjs/interview-tracker
![telegram-cloud-photo-size-2-5312557168999916960-x](https://github.com/user-attachments/assets/b65f3db3-da23-4652-a2ae-49d5c9659eb3)

OS: Windows

illright:
The issue is due to the unusual `tsconfig.json`:
```json
{
  "files": [],
  "references": [
    {
      "path": "./tsconfig.node.json"
    },
    {
      "path": "./tsconfig.app.json"
    }
  ],
}
```

The `@` alias is defined in `./tsconfig.app.json`, and it's not picked up by `tsconfck`, the library that we use for parsing tsconfigs.

PR #83, opened 2024-08-15
Widgets are falsely marked as not referenced

Desteel:
Resolves [78](https://github.com/feature-sliced/steiger/issues/78)

`tsconfck` [only handles](https://github.com/dominikg/tsconfck/discussions/186) the first level of `referenced` (and perhaps it won't be changed).

Anyway, it is already enough to cover general cases, like:

- [vite](https://github.com/feature-sliced/steiger/issues/78) tsconfig's `referenced`;
- monorepo, assuming that the steiger is run for a specific subproject;
- monorepo, assuming that only the first level of `referenced` is used (which is expected)

In the future, if the `tsconfck` behavior is changed, it will be used by default.


illright (changes requested):
Great PR, I just don't think we should make this functionality exclusive to the FSD plugin of Steiger :D How about we make these changes in `@feature-sliced/filesystem` instead?

illright:
This is a comment

packages/steiger-plugin-fsd/src/_lib/collect-related-ts-configs.ts
```diff
@@ -0,0 +1,75 @@
+import { TSConfckParseResult } from 'tsconfck'
+
+export type CollectRelatedTsConfigsPayload = {
```
illright, line 3: this is great

illright (dismissed):
Actually, I changed my mind, this function doesn't belong in the `@feature-sliced/filesystem` package. It doesn't belong to the FSD plugin either, but a better way to implement it needs some more thinking, so it will probably land in the next release of Steiger (see https://github.com/feature-sliced/steiger/discussions/107).

Thanks for this PR! Sorry it took so long :D

illright (approved):

PR #88, opened 2024-08-20
Create the toolkit for plugin authors

### LINKED ISSUE #71: Implement the type safety utilities for plugin authors
- [x] Make sure all config type definitions (`Config`, `ConfigObject`, `Plugin`, `Severity`) are exported
- [x] Make sure that the FSD plugin uses them to export a strongly typed object
- [x] Implement everything in `import { createPlugin, enableAllRules, type ConfigObjectOf } from '@steiger/toolkit'`
- [x] Implement `createConfigs` similarly to `createPlugin` 

---

Test cases:
- [x] An end-user of the linter should be able to import the FSD plugin and the `FSDConfigObject` and use `satisfies` to gain autocompletion for the available rules
- [x] An end-user can quickly create a custom plugin using the `createPlugin` function to get auto-completion for plugin fields and obtain a highly typed plugin object in the end
- [x] An end-user can define an object of configs containing the `all` and a hand-picked custom config with the `createConfigs` function and get auto-completion on the rule names
- [x] A plugin writer can use the `ConfigObjectOf` to export a type object for their plugin

illright:
Closes #71 

This is a significant refactoring to the types and locations of utilities. Few things to note:
- This PR adds a new package, `@steiger/toolkit`. It has these exports:
  ```ts
  export type * from '@steiger/types'
  
  export { findAllRecursively } from './find-all-recursively.js'
  export { enableAllRules, createConfigs } from './create-configs.js'
  export { createPlugin } from './create-plugin.js'
  export type { ConfigObjectOf } from './config-object-of.js'
  
  export { compareMessages, createFsMocks, joinFromRoot, parseIntoFolder } from './prepare-test.js'
  ```
  With it, the definition of the FSD plugin for Steiger looks like this:
  ```ts
  const plugin = createPlugin({
    meta: {
      name: 'steiger-plugin-fsd',
      version: packageJson.version,
    },
    ruleDefinitions: [/* ... */],
  })
  
  const configs = createConfigs({
    recommended: enableAllRules(plugin),
  })
  
  export default {
    plugin,
    configs,
  }

  export type FSDConfigObject = ConfigObjectOf<typeof plugin>
  ```
- I also fixed some type errors in Steiger core
- The tests for these type utilities are currently missing. I tried to make them, but it's a whole separate story, and often you fight the testing framework more than actually testing. We will consider this in the future.

Issues to fix:
- [ ] Context typings are beyond non-ergonomic, you have to type out the entire array of rules in the template args

Fixing it is beyond my power right now, I would like to at least get the baseline down. We don't use context currently, so once we start using it, then we can improve on this.


illright:
@daniilsapa could you review this please?

daniilsapa:
@illright 

Great job!
Most likely, we’ll need to add more utils in the future but you've prepared an excellent baseline.
I like that it has a solid foundation that covers 4 purposes:
- Creating a plugin
- Testing rules for a plugin
- Creating configuration sets for a plugin
- Utilities for custom rule code (`findAllRecursively` so far)

I really don't have any suggestions or complaints about it. I think it's good to go. 👍

Should we open an issue to create a short guide on how to build plugins for Steiger, so we remember that this task is still on the plate? Also, we could open an issue to cover the toolkit with tests as you mentioned that was a whole separate story.

illright:
Yep, I opened an issue for type tests already (https://github.com/feature-sliced/steiger/issues/98), and I will also create an issue for the guide. I still don't fully know where is best to host these guides. For 1.0.0, we will definitely want a dedicated docs website, but for now we might use GitHub wiki.

illright:
I'll let you merge your PR first and then take care of the conflicts and merge this one after

packages/toolkit/tsup.config.ts
```diff
@@ -0,0 +1,15 @@
...
+  entry: ['src/index.ts'],
+  format: ['esm'],
+  dts: {
+    entry: 'src/index.ts',
+    resolve: true,
+  },
+  treeshake: true,
+  clean: true,
+  esbuildOptions(options) {
+    options.define = { 'import.meta.vitest': 'undefined' }
```
daniilsapa, line 13: By the way, I've been meaning to ask you for a while. Why do some functions have tests for them placed in the same file? Why can't we just put them in separate files and remove that redundancy of having to replace `import.meta.vitest` with undefined?

illright: I just find it more convenient sometimes, especially when there's not a lot of code and not a lot of test cases. Having a separate `*.spec.ts` gets in the way of the file structure, so if I have a couple files, the duplicated names of test files are a bit annoying to deal with.

illright:
Okay, I merged the updates and cleaned everything up. Also tested it with the test cases from #71. @daniilsapa could you give it a final review please?

packages/steiger/src/app.ts
```diff
@@ -65,7 +65,3 @@ export const linter = {
     return [runRulesFx.doneData, () => watcher.close()] as const
   },
 }
-
-export function defineConfig(config: Config) {
```
daniilsapa, line 69: Great updates!

I have only one question about this thing.
 
Could you please explain why you decided to remove this function? I know it just returned the same object you passed it. This would cause errors like "The requested module does not provide an export named 'defineConfig'" and could be annoying for the users. And I know that the project is in the beta stage, so they should expect some APIs to change. 

But in the future, it might happen that we need to do some pre-processing on the config. In fact, I was thinking the opposite, about adding some flag like "defined" for the config that went through `defineConfig` and considering it the only valid config. Yeah, and throw an exception if it doesn't have the flag, so people get used to wrapping their configs in `defineConfig` for the sake of some far-off new but quite possible features added to `defineConfig`.

What do you think?

illright: Actually, I think that I removed it by mistake, thanks for bringing that up. I wanted to beef up the type inference capabilities of this function, but gave up for the moment, and then just accidentally removed the whole thing.

> But in the future, it might happen that we need to do some pre-processing on the config.

Nah, I would really prefer the `defineConfig` function to just be for typing. That seems to be the convention with other tools in the JS ecosystem, no one expects this function to be a requirement or to modify the config in any way. Besides, any processing we could do through this function could also be done in Steiger itself, without putting any burden on the user

I'll bring this function back

daniilsapa (approved):
I don't see any issues now. Thanks!

PR #95, opened 2024-08-23
Implement the complete solution for processing the flat configuration structure

### LINKED ISSUE #82: Implement the inclusion/exclusion of files in the config
## Specification
Here's my proposed mental model — since we run rules individually, let's consider a single rule at a time. For a rule, there are several cases:
1. **The rule is off for the entire codebase**  
   The solution here is simple — just don't run it
2. **The rule is off for a certain portion or portions A of a codebase, and enabled as `error` for the rest of the codebase**
   We create a clone of the VFS that excludes these portions A of the codebase and run the rule on the result. It will be as if these files simply don't exist. For the purposes of FSD, these files will be treated as "outside of the FSD root", so any references to these files from the FSD root are unregulated, and any references of FSD slices in these excluded files are invisible to the rules.
3. **The rule is off for a certain portion or portions A of a codebase, enabled as `warn` for different non-overlapping portions B, and then enabled as `error` for yet different non-overlapping portions C** (see diagram)  
   We create a clone of the VFS that includes B + C and run the rule on the result. Then we _somehow_* determine what the severity is

image

\* About determining severity. This is a difficult question, because some errors are about entire folders, and the config allows us to change severity of individual files. I suggest we make the `error` severity "viral", in the sense that if at least one file in the mentioned folder has severity `error`, then that diagnostic will have severity `error`, same for `warn`, going further down.

### 1. Config merging

First of all, I think all rule definitions (i.e. plugin array elements that add new rules) should be hoisted to the top, so the order should not prevent you from disabling a rule. For example, the following two configs should be identical:
```ts
export default [
  { rules: { 'fsd/public-api': 'error' } },
  { paths: 'src/shared/**', rules: { 'fsd/public-api': 'warn' } },
  fsd.plugin, // here it is last
]
```

```ts
export default [
  fsd.plugin, // here it is first
  { rules: { 'fsd/public-api': 'error' } },
  { paths: 'src/shared/**', rules: { 'fsd/public-api': 'warn' } },
]
```

With that out of the way, I propose that the result of merging a config array is a mapping of rule to a pair of VFSes, one for error, other for warn, where they are enabled. Initially it will be `new Map(rules.map(rule => [rule, { warn: null, error: initialVfs }]))`, but as the array elements are parsed, these VFSes should be updated individually for each rule. 

> [!NOTE]
> We can ignore performance for now, since at the start, there won't be that many rules, and the configs will probably not be too complicated. Later, we can profile and determine what should be optimized.

### 2. Running rules

When we run a rule, we have two VFSes (`{ warn, error }`), so ideally we should also implement a merge operator for VFSes. Then we merge the two VFSes, run the rule, get a list of diagnostics and determine the severity of each diagnostic with the algorithm proposed above

daniilsapa:
Resolves #82 

Hi @illright 

I know it's huge, but it didn't make any sense to partition this work, because it was kind of bringing all that we've done together. 

To make it a little easier for you to review, I've created a diagram of the config parsing and application flow:
(you can find types that the processes pass to each other in `steiger/src/models/config/types.ts`)

Screenshot 2024-08-23 at 14 13 05

### Notes

The function that I wrote for glob matching did not fully suit our needs because it removed nodes from the tree. And as I saw we needed more of a thing that would mark nodes instead of removing them because we’ll need to run globs on a tree several times e.g. for the following config. We need to disable /shared first and then re-enable it (yeah, we discussed that it’s a weird config, but it still could be specified). So, I decided that we needed to mark/remark nodes for every set of globs provided for a rule, and only when all sets are processed we can safely remove nodes that remained “off” 

```javascript
const config = [
    ...fsd.configs.recommended,
    {
      files: ['./src/shared/**'],
      rules: {
        'fsd/public-api': 'off',
      },
    },
    {
      files: ['./src/shared/**'],
      rules: {
        'fsd/public-api': 'error',
      },
    },
],
```

I installed `ramda` library. It has a lot of cool functional utils. One of the most useful functions is pipe which allows you to run operations on an array with one traversal, like filtering and mapping in the example below. There aren't many places where it's used now, but we can keep adding more in the future. It should improve the performance a bit by removing the need to re-traverse arrays with chaining array methods

Screenshot 2024-08-23 at 14 27 18


packages/types/index.ts
```diff
@@ -67,7 +67,7 @@ export interface ConfigObject {
   /** Globs of files to ignore */
   ignores?: Array<string>
   /** Severity of rules and individual rule options. */
-  rules?: {
+  rules: {
```
illright, line 70: issue: omitting `rules` is a way of saying "apply this to all rules. It's made to let users quickly ignore certain parts of their codebase like tests or compiled files

suggestion: let's make this key optional and add support for these cases

packages/steiger/src/models/config/process-configuration/create-rule-instructions.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      {
+        rules: {
+          rule1: ['warn', {}],
+          rule2: ['error', {}],
+        },
+      },
+      {
+        files: ['src/shared/ui/**/*', 'src/entities/user/ui/**/*'],
+        rules: {
+          rule1: ['warn', {}],
```
illright, line 106: issue: I think we shouldn't throw errors when the specified options are the same. I believe that it's not immediately obvious what's the correct way of setting different severities while retaining the same options (even though we will guide users with error messages), and I anticipate that some people will want to simply duplicate the options object rather than omitting it, and since such a config doesn't contain any contradictions, we shouldn't throw errors at the user in this case.

I just think that running up against an error message is always frustrating for the user, so we should always try to minimize the error surface

suggestion: let's adjust this test to check for not throwing, then add another test that checks that an error will be thrown if two different options are specified

illright (changes requested):
Thanks for this! I haven't finished reviewing yet, I just wanted to drop some early comments about some runtime specifics that I think are important. I'm also not sure I understand what a `RuleRunEnvironment` is, could you add a docstring that describes what this object represents?

packages/types/index.ts
daniilsapa: I would be happy to but if we don't have `rules` in a config object, it means that we have only globs, and then there is nothing to apply. 

We could think up an alternative syntax or extend the object shape for that purpose.

E.g. I thought of the following solutions to the issue

1. You still need to specify the `rules` object but can apply severities to all rules
```javascript
export default defineConfig([
  ...fsd.configs.recommended,
  {
    files: ['**/shared/**'],
    rules: {
      '*': 'off' // can only set severity in this case (options obviously cannot be set)
    }
  }
])
```

2. Additional "severity" field that is processed only when the rules object is absent (otherwise it's skipped). But this slightly contradicts with the declared policy of keeping the public interface as concise as possible.
```javascript
export default defineConfig([
  ...fsd.configs.recommended,
  {
    files: ['**/shared/**'],
    severity: 'off'
  }
])
```

packages/steiger/src/models/config/process-configuration/create-rule-instructions.test.ts
daniilsapa: 100%, that's a good point. I'll make the fixes

packages/types/index.ts
illright: Ah, you're right, I didn't think this through. I just kind of copied ESLint here without really thinking if it makes sense or not.

In ESLint, there's a [special kind of configuration object](https://eslint.org/docs/latest/use/configure/configuration-files#globally-ignoring-files-with-ignores) — if you only specify the `ignores` key and nothing else, that will be taken to mean "ignore these files for all rules". That's not how `ignores` behaves when you have `files` or `rules` specified, but I think I'd prefer consistency with ESLint here over creating a new syntax. 

So I think that our `ConfigObject` should be `{ files?, ignores?, rules } | { ignores }`, so then we don't make `rules` optional, but still leave an opportunity to only have `ignores` and nothing else

daniilsapa: No problem, I can implement logic handling that `{ ignores: [] }` case.

There’s no doubt that ESLint is an amazing high-quality tool, one of the best of its type. I have been using it myself for many years. And since we're developing a tool of the same type, references to ESLint are 100% reasonable, and making the config structure similar to it should allow people to onboard to Steiger quicker and smoother.

I just think that we would like to be sure that we consider what's the best/more convenient for users in our particular case in the first place and we don't just blindly copy and paste from ESLint/Stylelint/… whatever.

illright (commented):
I also have one high-level architectural concern for this PR. It introduces a lot of functionality and responsibility to the `config` model, and it's hard for me to wrap my head around it all. I'd prefer if the config model was only responsible for reading the config and answering questions about it in a format that's convenient for Steiger itself. Then the logic of applying globs could be extracted to a feature, like it used to be before this PR.

I tried to imagine the flow of communication that would keep the config model as slim as possible, and here's what I came up with. This diagram is an imaginary dialog between Steiger (the `app.ts` file) and the `config` model, going from top to bottom. Underlined are the inputs that Steiger supplies to the queries/commands to the config:

image

However, you've done some great work here, so ideally we would come to a solution somewhere in the middle. Do you think that what I've described a good idea or would you prefer to keep the code as it is?

packages/steiger/src/app.ts
```diff
@@ -34,35 +27,24 @@ const $enabledRules = combine($config, $rules, (config, rules) => {
...
+      }
 
-      return Promise.resolve(rule.check(vfs, optionsForCurrentRule)).then(({ diagnostics }) =>
+      return Promise.resolve(rule.check(ruleVfs, optionsForCurrentRule)).then(({ diagnostics }) =>
         diagnostics.map<AugmentedDiagnostic>((d) => ({
           ...d,
           ruleName: rule.name,
           getRuleDescriptionUrl,
-          severity: severities[rule.name],
+          severity: ruleEnv.severityMap[d.location.path],
```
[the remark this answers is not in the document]
daniilsapa: Good catch! Folders are not in the severity map, I'll add them

packages/steiger/src/app.ts
```diff
@@ -1,29 +1,22 @@
...
+import { prepareRuleRunEnvs } from './models/config'
+import { RuleRunEnvironment } from './models/config/types'
 
 function getRuleDescriptionUrl(ruleName: string) {
   return new URL(`https://github.com/feature-sliced/steiger/tree/master/packages/steiger-plugin-fsd/src/${ruleName}`)
 }
 
-type Config = typeof $config
-type SeverityMap = Record<string, Exclude<Severity, 'off'>>
+type Config = typeof $ruleInstructions
```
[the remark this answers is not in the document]
daniilsapa: Yes, I'll rename it

packages/steiger/src/app.ts
```diff
@@ -34,35 +27,24 @@ const $enabledRules = combine($config, $rules, (config, rules) => {
   )
```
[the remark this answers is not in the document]
daniilsapa: Yeah, looks like I didn't pay enough attention to this feature. I'm going to refactor it.

packages/steiger/src/models/config/prepare-rule-run-envs/mark-severities.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export default function markSeverities(
+  ruleToInstructions: Record<string, RuleInstructions>,
+  ruleToVfs: Record<string, Array<File>>,
+): Record<string, Array<SeverityMarkedFile>> {
+  const ruleToMarkedFiles: Record<string, Array<SeverityMarkedFile>> = Object.entries(ruleToVfs).reduce(
+    (acc, [ruleName, initialVfs]) => {
+      return {
+        ...acc,
+        [ruleName]: initialVfs.map(markDefault),
```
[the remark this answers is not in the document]
daniilsapa: Hmm, yes, this may lead to unexpected situations, I think it's better to change it to "off".

daniilsapa:
I agree with your concern, the config model has become too bloated with responsibilities and code in general. I will try to fix this problem tomorrow, and yes, your diagram offers an interesting idea and will come in handy, thanks!

daniilsapa:
Hi @illright 
I fixed all your minor call outs. Also, I addressed your architectural concern about the config model

illright (changes requested):
Thanks for the changes, it is indeed better. However, I still feel like the solution is more complicated that I'd like it to be, which might impact maintainability in the long run. The concrete points of complexity, in my opinion, are:
1. The `apply-severity-globs-to-vfs` feature. From the perspective of Steiger's business domain, and to use the FSD understanding of a _feature_, this isn't a feature because it doesn't do anything meaningful to the user by itself. That's why I find this feature and its abstractions `VfsWithSeverity` and `SeverityMarked*` hard to understand.
2. The flat representation of the VFS (record of path to VFS object). The fact that there are several representations for the same thing makes it hard for me to understand what parts of code are compatible with what other parts.
3. Stepping out of Effector philosophy in certain places. I noticed that in some places in code we have `.getState()` calls and getter functions, and I believe that Effector is meant to work a bit differently, instead with derived stores and meaningful events (for example, "load this config") and effects.
4. Consistency of error production and handling. Currently, errors are thrown in many places, for example, when building a validation schema, when creating rule instructions, when accessing some of the getters, when parsing a config with Zod. I'm getting lost in these places because they also include human-readable error messages, which ideally I'd prefer to keep all in one place, somewhere higher up, like in `app.ts`, to keep a close look over the amount, variety, and clarity of possible error messages. There are also ways to handle errors in Effector using effects, which might also be a good solution if we isolate this functionality well.

I really wish we had a more real-time communication channel than simply back-and-forth with reviews. I can think of Telegram, Discord, or even maybe a video call to do some brainstorming. What do you think?

daniilsapa:
@illright 

1. I agree that it's not a feature from Steiger's business domain perspective. So, I'll consider moving it to another place.
I described the need for severity-marked FS entities here:
> The function that I wrote for glob matching did not fully suit our needs because it removed nodes from the tree. And as I saw we needed more of a thing that would mark nodes instead of removing them because we’ll need to run globs on a tree several times e.g. for the following config. We need to disable /shared first and then re-enable it (yeah, we discussed that it’s a weird config, but it still could be specified). So, I decided that we needed to mark/remark nodes for every set of globs provided for a rule, and only when all sets are processed we can safely remove nodes that remained “off”

```javascript
const config = [
    ...fsd.configs.recommended,
    {
      files: ['./src/shared/**'],
      rules: {
        'fsd/public-api': 'off',
      },
    },
    {
      files: ['./src/shared/**'],
      rules: {
        'fsd/public-api': 'error',
      },
    },
],
```

2. The flat representation of VFS is used only inside `apply-severity-globs-to-vfs` because it's convenient to present it in that way. But it consumes and produces a tree-shaped VFS, so you don't need to think of different representation outside that one module
3. Got it, I'll be delving deeper into Effector (I just heard about it and had never used it before this project) to make some parts fit better with its philosophy.
4. Agree, I'll refactor it, so the errors are thrown in a more expected manner

And yes, I think it would significantly increase productivity and improve coordination. In fact, I was thinking of suggesting the same. Telegram would be the most convenient option. I will send you an invite on Linkedin and share my contact.

> I really wish we had a more real-time communication channel than simply back-and-forth with reviews. I can think of Telegram, Discord, or even maybe a video call to do some brainstorming. What do you think?

daniilsapa:
Hi @illright 

Here are some notes on the latest changes:

- I could not implement something like this `const { off: offGlobs, warn: warnGlobs } = getGlobsForRule(rule.name)` that we discussed on the call because we cannot group globs by severity as later ones may override the severity of earlier ones, so they need to be run exactly in the sequence they appear in the config.
- I implemented the new behavior of GlobalIgnores.
- I didn’t understand any reason why AugmentedDiagnostic needs to have the getRuleDescriptionUrl method and decided to replace it with a simple string field because we can get the URL right away.
- I renamed Diagnostic to PartialDiagnostic and renamed AugmentedDiagnostic to just Diagnostic. (We discussed that briefly during the call). Also, I moved I moved diagnostic from pretty-reporter to @steiger/types.
- It turned out that relative paths in globs were not supported (`files: ["./src/shared/**"]`), so I added support for them

illright:
Looking over through a few files from my phone, I can already answer the question about `getRuleDescriptionUrl`. The way it's done in your PR now will only work for the FSD ruleset, not any others, if I understand correctly, because the URL to `feature-sliced` is hardcoded in Steiger core.

The reason why I wanted to make it a function is to allow plugin writers to provide their own logic of generating description URLs. However, I now realize that I haven't thought through properly where this function will actually be supplied. I think we should make another field on the plugin object called `getRuleDescriptionUrl`, and Steiger core should pull that function from there.

daniilsapa:
> Looking through a few files from my phone, I can already answer the question about getRuleDescriptionUrl. The way it's done in your PR now will only work for the FSD ruleset, not any others, if I understand correctly, because the URL to feature-sliced is hardcoded in Steiger core.

Oh yeah, I sometimes forget that feature-sliced ​​doesn't have to be the only plugin. That can be harmful, so I'll make a note to remind myself about custom plugins before I start my working session on Steiger.


> The reason why I wanted to make it a function is to allow plugin writers to provide their own logic of generating description URLs. However, I now realize that I haven't thought through properly where this function will actually be supplied. I think we should make another field on the plugin object called getRuleDescriptionUrl, and Steiger core should pull that function from there.

Yes, that makes sense, and it looks like a task for a separate issue. Should I revert the change with `getRuleDescriptionUrl` or should I leave it as is if we're going to develop a full-fledged solution anyway.

illright (changes requested):
Phew, okay, finished reviewing. I definitely like the direction in which this is heading.

By the way, could you avoid force-pushing in the future? The reason is because it messes with GitHub's tracking of which files were reviewed, which means that I'm no longer sure which files I've seen and which files I haven't. The commit history of the PR will be squashed anyway, so that shouldn't be too much of a concern

illright:
> Should I revert the change with getRuleDescriptionUrl or should I leave it as is if we're going to develop a full-fledged solution anyway.

If it isn't too much work, I'd prefer if we reverted that part, yes. Simply less things to track while thinking about this PR. If it's too complicated, then yeah, we can leave it as it is

packages/steiger/src/app.ts
```diff
@@ -1,87 +1,40 @@
...
-      )
-    }),
-  )
-  return ruleResults.flat()
+import { $enabledRules, getEnabledRules, getGlobalIgnores } from './models/config'
+import complementDiagnostics from './features/complement-diagnostics'
+import { runRule } from './features/run-rule'
+import removeGlobalIgnoresFromVfs from './features/remove-global-ignores-from-vfs'
+
+async function collectDiagnostics({ vfs, rules }: { vfs: Folder; rules: Array<Rule> }) {
```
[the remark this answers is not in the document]
daniilsapa: Yeah, no problem

packages/steiger/src/features/complement-diagnostics/index.ts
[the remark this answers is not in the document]
daniilsapa: Overall this makes sense, but I wasn't sure about the name because `assign-{something}` sounds like a function that modifies some object passed to it to add a field. So, if we are just going to return severity I decided to go with calculate-severity so far.

packages/steiger/src/features/remove-global-ignores-from-vfs/index.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import removeGlobalIgnoresFromVfs from './index'
+import { joinFromRoot } from '../../_lib/prepare-test'
+
+describe('removeGlobalIgnoresFromVfs', () => {
+  it('should remove nodes that match global ignores from VFS', () => {
+    const vfs = {
+      type: 'folder',
+      path: joinFromRoot('src'),
+      children: [
```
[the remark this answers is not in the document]
daniilsapa: 100%

packages/steiger/src/models/config/create-rule-instructions.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export default function createRuleInstructions(
+  config: Config,
+  configLocationPath: string,
+): Record<string, RuleInstructions> {
+  const ruleNameToInstructions: Record<string, RuleInstructions> = preCreateRuleInstructions(config)
+
+  return config.reduce((acc: Record<string, RuleInstructions>, item) => {
+    if (isConfigObject(item)) {
+      Object.entries(item.rules!).forEach(
```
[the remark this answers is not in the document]
daniilsapa: Yes, it's not necessary, looks like it left over from the previous iterations

packages/steiger/src/features/complement-diagnostics/index.ts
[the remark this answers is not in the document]
illright: That's true. `calculate-severity` is also good

packages/steiger/src/models/config/build-validation-scheme.spec.ts
```diff
@@ -0,0 +1,311 @@
+import { describe, expect, it } from 'vitest'
+import { Config, Plugin } from '@steiger/types'
+
+import buildValidationScheme from './build-validation-scheme'
+import { isPlugin } from './raw-config'
+
+function checkToAnyFunction(config: Config) {
```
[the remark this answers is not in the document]
daniilsapa: Yes, it's not intuitive at all, I forgot to drop a comment for it

packages/steiger/src/models/config/build-validation-scheme.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+    const config = [dummyPlugin] as Config
+    const scheme = buildValidationScheme(config)
+
+    expect(() => scheme.parse(config)).toThrow('At least one config object must be provided!')
+  })
+
+  it('should throw an error if a config object without rules is provided', () => {
+    const config = [dummyPlugin, {}] as Config
+    const config1 = [dummyPlugin, { files: ['/shared'] }] as Config
+    const config2 = [dummyPlugin, { ignores: ['**/*.test.ts'] }] as Config
```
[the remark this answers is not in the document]
daniilsapa: In this case `dummyPlugin` contains only the plugin object itself without any config objects, so:
1. `{}` is not allowed because only 3 shapes of objects are allowed in a config (Plugin, ConfigObject, GlobalIgnore)
2. This shape `{ files: ['/shared'] }` is not valid because it's not a plugin, global ignore, or config object.
3. `{ ignores: ['**/*.test.ts'] }`. Even though it's a valid global ignore, this config still misses a single config object, which is nonsense, as no rules are configured to be run

packages/steiger/src/models/config/create-rule-instructions.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { describe, expect, it } from 'vitest'
+import createRuleInstructions from './create-rule-instructions'
+import { Config } from '@steiger/types'
+
+describe('createRuleInstructions', () => {
```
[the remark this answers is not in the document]
daniilsapa: Yes, makes sense

packages/steiger/src/models/config/create-rule-instructions.ts
```diff
@@ -0,0 +1,73 @@
...
+function convertRelativeGlobsToAbsolute(rootPath: string, globs: Array<string>) {
+  function composeAbsolutePath(root: string, glob: string) {
+    // Remove '/'. The root has platform-specific separators
+    const segmentsOfRoot = root.slice(1).split(sep)
+
+    // Remove './' from the beginning of the glob. Globs always have '/' as separators
+    const segmentsOfGlob = glob.slice(2).split('/')
+
+    return `/${[...segmentsOfRoot, ...segmentsOfGlob].join('/')}`
+  }
```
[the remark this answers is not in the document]
daniilsapa: Yep, makes sense

packages/steiger/src/cli.ts
```diff
@@ -56,8 +56,9 @@ const { config, filepath } = (await cosmiconfig('steiger').search()) ?? { config
 const defaultConfig = fsd.configs.recommended
 
 try {
+  const configLocationDirectory = filepath ? filepath.split(sep).slice(0, -1).join(sep) : ''
```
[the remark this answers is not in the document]
daniilsapa: Yes, it's better to fix that

daniilsapa:
@illright 
Looks like I fixed the callouts. Also, I reverted my changes for getRuleDescriptionUrl.

> By the way, could you avoid force-pushing in the future? The reason is because it messes with GitHub's tracking of which files were reviewed, which means that I'm no longer sure which files I've seen and which files I haven't. The commit history of the PR will be squashed anyway, so that shouldn't be too much of a concern.

Ok, I only did that because I needed to generate a new change log, as my new changes touched more packages. So, I dropped the previous change log commit using an interactive rebase and created a new one. But now I think that I could just create a new change log that overrides the previous one, or somehow complement the previous one.


> suggestion: I'm not sure yet, but I think we may want to encode a rule that if all files from a folder are excluded, then the folder itself is excluded too. The reason is that I wouldn't want users to duplicate their globs like that

Yes, maybe. It's a significant difference between ESLint, Stylelint, and Steiger - folders. Other linters don't have to care about reporting errors/warnings on folders.
But, at the end of the day, I think these duplications should not be a problem, because usually rules that check folders and files are different rules, so in the real world you will rarely encounter such cases in one glob array:

```
const config = [
  { ignores: ['**/__mocks__', '**/__mocks__/**'] }
]
```

illright:
> so in the real world you will rarely encounter such cases in one glob array

I think the problem is the convention, what users are used to. If you write `**/__mocks__/**` for one rule, you're more likely to do the same for another rule (because you cannot tell without looking at the code if it matters or not). Besides, I think that our FS discovery algorithm also works to not add folders if they don't have anything inside

illright (changes requested):
Ooooh, we're so close! The comments I left are not that big of a deal, but I think we should fix that glob thing so that `__mocks__/**` also removes the folder

packages/steiger/src/features/run-rule/calc-and-remove-off-nodes.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+describe('calcAndRemoveOffNodes', () => {
+  it("should remove off nodes when there's an off glob group", () => {
+    const globs = [
+      { severity: 'error', files: [], ignores: [] },
+      { severity: 'off', files: ['**/shared', '**/shared/**'], ignores: [] },
+    ] as Array<GlobGroup>
+
+    const vfs = parseIntoFsdRoot(
+      `
+      📂 src
```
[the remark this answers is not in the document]
daniilsapa: Oh, yes

packages/steiger/src/features/calculate-severity/index.ts
```diff
@@ -0,0 +1,29 @@
+import { Severity } from '@steiger/types'
+import { getGlobsForRule, GlobGroup } from '../../models/config'
+import { createFilterAccordingToGlobs } from '../../shared/globs'
+
+function getFinalSeverity(path: string, globGroups: Array<GlobGroup>): Exclude<Severity, 'off'> {
+  let finalSeverity: Severity = 'error'
```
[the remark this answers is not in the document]
daniilsapa: Because in this case the severity either won't change or will change to `warn`. This feature runs when `off` nodes are removed.

packages/steiger/src/models/config/create-rule-instructions.ts
```diff
@@ -0,0 +1,75 @@
...
+
+function convertRelativeGlobsToAbsolute(rootPath: string, globs: Array<string>) {
+  function composeAbsolutePath(root: string, glob: string) {
+    // Remove '/'. The root has platform-specific separators
+    const segmentsOfRoot = root.slice(1).split(sep)
+
+    // Remove './' from the beginning of the glob. Globs always have '/' as separators
+    const segmentsOfGlob = glob.slice(2).split(posix.sep)
+
+    return `/${posix.join(...segmentsOfRoot, ...segmentsOfGlob)}`
```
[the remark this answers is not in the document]
daniilsapa: Fixed

packages/steiger/src/models/config/create-rule-instructions.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      },
+      {
+        files: ['src/entities/user/ui/**/*'],
+        rules: {
+          rule1: 'error',
+        },
+      },
+    ]
+
+    expect(createRuleInstructions(config, '')).toEqual({
```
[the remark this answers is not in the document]
daniilsapa: Fixed

daniilsapa:
I implemented the desired behavior for globs but had to change the shape of PartialDiagnostic a bit. It took a while but I thought that the change I made would be useful in other scenarios in the future.

Also, I noticed that relative globs in GlobalIgnore objects were not converted, so I added logic to convert them too.

packages/types/index.ts
```diff
@@ -18,20 +18,27 @@ export interface Rule<Context = void, RuleOptions = BaseRuleOptions> {
...
 }
 
-export interface Diagnostic {
+export interface PartialDiagnostic {
   message: string
   fixes?: Array<Fix>
   location: {
     /** Absolute path to a folder or a file that contains the issue. */
     path: string
+    type: 'file' | 'folder'
```
illright, line 30: issue: this addition is kind of a bummer, I wouldn't want to hurt the plugin authors' experience for that change in the matching logic. Besides, I don't think this is necessary, since the code checks if the glob ends with `**`, we know that this cannot be a file, it has to be a folder.

suggestion: let's roll this one back

packages/steiger/src/shared/globs.ts
```diff
@@ -1,3 +1,67 @@
...
+    return defaultResult
+  }
+
+  // If the default result failed, try to match the containing folder
+  if (nodeType === 'folder' && isAllInsideFolderPattern) {
+    return minimatch(path, pattern.slice(0, -3))
+  }
+
+  return false
+}
```
illright, line 36: issue: I think there are edge cases that this logic doesn't cover. Ideally, we'd want to work like Git — if there are no files in a folder, the folder doesn't exist. The edge case here would be like this:
- 📂 folder1
  - 📄 index.ts
  
and an ignoring glob of `**/index.ts`. This glob would leave folder1 empty, and this logic will not catch that.

suggestion: let's instead change the logic of the function that recomposes a flat VFS back into a tree to do a pass at the end that would remove all empty folders. That should be more robust and shouldn't require us to add any fields to the diagnostic.

packages/steiger/src/models/config/convert-relative-globs-to-absolute.ts
```diff
@@ -0,0 +1,41 @@
...
+    // Remove '/'. The root has platform-specific separators
+    const segmentsOfRoot = root.slice(1).split(sep)
+
+    return `/${posix.join(...segmentsOfRoot, glob)}`
+  }
+
+  return globs.map((glob) => (glob.startsWith('.') ? composeAbsolutePath(rootPath, glob) : glob))
+}
+
+export function convertRelativeGlobsToAbsolute(config: Config, configLocationPath: string | null) {
```
illright, line 17: praise: great that you decided to extract it into a separate module! It's easier to maintain and test that way

illright (changes requested):
Everything else is great, except for that last change — the auto-removal of empty folders. I wrote more details below

packages/steiger/src/shared/globs.ts
daniilsapa: Do we need this tweaked behavior only when we remove nodes? Because my goal was to make all globs work in the same way in all cases, so they are predictable. So, when we apply error/warn severities it matches the containing folder too, and that is why I put that logic on such a low level

packages/types/index.ts
daniilsapa: As I said, I thought it would be useful in the future for other functionalities, e.g.:
- for pretty-printer to print some icons of a file/folder for clarity and faster visual perception of the report
- to filter what diagnostics to display in IDEs (obviously, in an IDE, we can highlight errors only inside files)
- other sorting, rearranging, and partitioning operations on diagnostics
- etc.

And I thought it would be better to initially define which FS node a diagnostic applies to, rather than having to determine it in many places where diagnostics are processed.

So, are you sure we won't need that field in the future? Because if we see many applications for that field in the future, isn't it better to let plugin creators get used to it sooner than introduce a breaking change in the future?

packages/steiger/src/shared/globs.ts
daniilsapa: What I implemented was a kind of "local" containing folder matching where I matched it for a single glob. I had an idea of "global" matching and it was similar to what you proposed. The idea is to add another tree traversal at the end of severity processing to check if all nodes inside a folder have some type of severity and if it's true we mark the containing folder with the same severity. But it still has flaws as some parts of FS can be remarked, and that won't allow us to definitely say how to mark the containing folder

```javascript
// Here everything inside /lib is off except for get-query-params, which is 'warn'
const config = [
  {
    files: ['./src/shared/lib/**'],
    rules: {
       'fsd/forbidden-imports': 'off',
    }
  },
  {
    files: ['./src/shared/lib/get-query-params.js'],
    rules: {
       'fsd/forbidden-imports': 'warn',
    }
  }
]
```

daniilsapa:
@illright 
Phew, I found more and more edge cases as I worked, so I had to rewrite some parts several times and change the solution significantly overall.

Main changes
- I realized that we must run rules on full VFS and then filter diagnostics based on their severities. That's because some disabled folders can contain folders with error/warn severities and that's why we cannot remove it entirely from the tree.
- As we cannot remove nodes, I had to introduce an inner severity type that had an additional 'excluded' value for technical purposes. That means that a node is marked as removed forever and its severity cannot be overridden with later config objects. The public severity type remained unchanged.
- I handled edge cases for all folder globs based on what we discussed + some minor additions that I thought made the best sense

I created a table that will help to understand how the folder globs are applied now. I also added descriptions for each implicit rule type (based on analyzed node types).


| **Glob**            | **Folder rules**                                                            | **File rules**                            | **Potential combined rules**                                             |
|---------------------|-----------------------------------------------------------------------------|-------------------------------------------|--------------------------------------------------------------------------|
| `**/folder-name`    | Applies to only that folder                                                 | Applies to files that are direct children | Applies to the direct children and the folder itself                     |
| `**/folder-name/`   | The same as above                                                           | The same as above                         | The same as above                                                        |
| `**/folder-name/*`  | The same as above                                                           | The same as above                         | The same as above                                                        |
| `**/folder-name/**` | Applies to the folder itself and all folders inside it nested on all levels | Applies to all files nested on all levels | Applies to the folder itself, and all files/folders nested on all levels |

New application rules
- Strip trailing slashes
- The severity of a folder is max severity of its children children
- **/folder-name applies severity to all files directly nested inside that folder and to the folder itself



Let me know if you need more explanations or corrections. Thank you in advance!

illright (changes requested):
Hey, yeah, that's a lot of changes indeed. I feel like it's a step backwards though, architecture-wise :(

Particularly I'm concerned about the new set of `features`. My brain just filters out the word "wizard" in "vfs-severity-wizard", because I can't give it any meaning, so the name of the feature in my mind essentially becomes "something for VFSes and severities". From that name, I cannot infer any possible contents of the feature. I'd like us to go back to the previous set of features, where one runs the rules and another calculates severities.

> I realized that we must run rules on full VFS and then filter diagnostics based on their severities. 

I don't think we can do that. Since certain rules might trigger errors in one file based on the existence of another, we must remove the ignored files from the VFS before we run the rule. There's no robust way to filter out diagnostics after the fact, I believe.

> That's because some disabled folders can contain folders with error/warn severities and that's why we cannot remove it entirely from the tree.

That's why I wanted to avoid assigning severities to folders altogether — there are many edge cases and issues in that. If we only assign severities to files, and only remove files, then a whole class of issues disappears. The folders then simply follow files — the severity of a folder is inferred as the maximum severity of a child. If the severity of the folder is `off` (as a consequence of all children files being `off`), then that folder is removed from the VFS.

> As we cannot remove nodes, I had to introduce an inner severity type that had an additional 'excluded' value for technical purposes.

This is also a complication that I'd like to avoid. As we introduce more data states and labels, the flow of familiar objects becomes obstructed, so it's harder to understand what the feature is actually doing.

daniilsapa:
@illright 

## 1
> Particularly I'm concerned about the new set of features. My brain just filters out the word "wizard" in "vfs-severity-wizard", because I can't give it any meaning, so the name of the feature in my mind essentially becomes "something for VFSes and severities". From that name, I cannot infer any possible contents of the feature. I'd like us to go back to the previous set of features, where one runs the rules and another calculates severities.

Ok, indeed, wizard can be a bit vague, so I’ll come up with a better name and most likely shape too.

## 2
> I don't think we can do that. Since certain rules might trigger errors in one file based on the existence of another, we must remove the ignored files from the VFS before we run the rule. There's no robust way to filter out diagnostics after the fact, I believe.

> If the severity of the folder is off (as a consequence of all children files being off), then that folder is removed from the VFS.

Ok, but what to do if we have the following case? Right now it says: "Mark all files that are direct children as off" (and subsequently the folder itself). But we cannot remove the folder completely as it can contain subfolders that have severities other than 'off'. And from this problem follows the need to introduce an additional internal type of severity.

```javascript
  const config = [
    ...fsd.configs.recommended,
    {
      files: ['./src/app'],
      rules: {
        'fsd/forbidden-imports': 'off'
      }
    }
  ]
```

## 3
> That's why I wanted to avoid assigning severities to folders altogether — there are many edge cases and issues in that. If we only assign severities to files and only ignore files, then a whole class of issues disappears. 

Yeah, I can make folder severity calculation fully dynamic but the issue I described in the 2nd section still persists

daniilsapa:
@illright 
Also, some folders do not have files by default according to FSD (usually these are layer folders `/pages`, `/widgets`, etc). So, there should be a way to calculate their severity too.

illright:
To address your concern 2: let's say we have the following structure in :
- 📂 app
  - 📂 entrypoint
  - 📂 styles
- 📂 pages
  - 📂 home

The glob `"./src/app"` matches nothing, because it's only targeting a file called `app` in the `./src` folder. To target the entire `app` folder, you should use `"./src/app/**"`.

We can offer guidance to people by analyzing the config at runtime and verifying that there are no cases when someone wanted to target the entire folder but instead targeted a non-existent file of the same name

illright:
> Also, some folders do not have files by default according to FSD

That's okay. The calculation of severity for a folder involves knowing the severities of direct children, whether they are files or folders. If they are files, they have a severity of their own. If they are folders, their severity is calculated recursively

daniilsapa:
> The glob "./src/app" matches nothing, because it's only targeting a file called app in the ./src folder. To target the entire app folder, you should use "./src/app/**".

Yes, it can work as that, but we still have another glob type `./src/app/*` that matches all files and folders that are direct children of "./src/app". 

> That's okay. The calculation of severity for a folder involves knowing the severities of direct children, whether they are files or folders. If they are files, they have a severity of their own. If they are folders, their severity is calculated recursively

Yes, but there's a possible issue with a recursive calculation. If we have just a single file e.g. "./src/entities/user/ui/UserAvatar.tsx" marked as "error" and we marked all other files as warn. The recursive propagation will make all the folders upstream marked as "error". Is that a correct behavior?

illright:
> we still have another glob type ./src/app/* that matches all files and folders that are direct children of "./src/app".

Yes, and if there's a folder ./src/app/entrypoint with at least one file, this glob will not be enough to mark the whole folder with the given severity, because globs only apply to files, not to folders

> Is that a correct behavior?

Yep

daniilsapa:
Hmm, okay, let me re-think it a bit, because there was a little misunderstanding of how this is supposed to work.

I'll think hard to find edge cases and possible weird moments for the end user tomorrow. If it gracefully comes together as a whole I'll just update the PR, or we'll discuss additionally if it falls apart somewhere.

Also, I just thought about an auxiliary tiny browser app where a user can add their folder and file structure using the File System API, we build a visual representation for their FS tree. Then the user can write some config objects and ignores to a form. Affected nodes will light up red for "error", yellow for "warn" and gray for "off". I just thought that if our glob syntax and usage would be a bit unusual, such a helper could help the users have visual feedback when they set up the globs in the config, understand it better and faster, and also debug problems. But I'm not sure if it's worth it. What do you think about this idea?

illright:
About the folder visualizer — it's definitely an interesting idea, and FSD itself would benefit from some kind of visualizer, but I agree that so far it's not clear if there's actually a need for this or not. Let's release the new config format and get some feedback, and if people complain that writing a config is too difficult, we can think about solutions, including this one

daniilsapa:
@illright 

I've brought back the previous set of features with a few differences.

Also, I renamed `vfs-severity-wizard` to `vfs-severity-map`. It's a reference to a physical map that would help you navigate through VFS and its severities to find what you want. I thought about naming it "atlas" to not confuse it with other programmatic concepts around map, but eventually decided to go with "map". I moved it to models as it's now used in several features. If you have a better name for that module, let me know.

I'm not sure that it should be in models, BTW, looks like we need to introduce some other level. We can discuss that.

Thank you in advance!

daniilsapa:
And yes, about the visualizer, looks like it could be highly reusable and be useful not just for globs but also in some examples in the docs and learning materials

daniilsapa:
Hi @illright 

I pushed my progress partially, could you take a look at /shared/globs when you have time? It looks like it’s aligned with your idea.   

P.S. I know that there are some cross-imports in /shared now.  I’ll consider moving some utils inside shared/globs if they are used only there or moving `applyExclusions` to another layer.

illright:
Yes, this is quite nice! One minor thing — could you please keep index files as simple re-exports, and move logic in new adjacent files? I usually use index files to quickly get an understanding of what a certain module is capable of, and having logic there gets in the way of that a bit.

daniilsapa:
Hi @illright 

A few notes on my changes:

- Yes, that requirement made perfect sense, so I brought the index file to the format where it just re-exports other files, and did that for several other modules too. (BTW, shouldn't we add such conventions to some document, so the contributors can start delivering code that follows them faster?)
- I got rid of `vfs-severity-map` completely. Indeed, after I implemented your idea, I could remove a lot of code and now the code base looks more straightforward and clean. I'm happy that you insisted on this idea, it's great 👍.
- I moved several file-system utilities (`removeEmptyFolders`, `copyNode`) to `/shared/globs` module. Because we used them only there, also it looks like applying a glob should be the only way to change an FS tree (I don't see any other scenarios so far).
- After I made my changes I noticed that `ramda` is used in just 2 places, so I decided to remove it.

packages/steiger/src/features/remove-global-ignores-from-vfs/remove-global-ignores-from-vfs.ts
```diff
@@ -0,0 +1,17 @@
+import { Folder, GlobalIgnore } from '@steiger/types'
+import { applyExclusion } from '../../shared/globs'
+import { mergeGlobGroups } from '../../shared/globs/merge-glob-groups'
+
+function turnGlobalIgnoresIntoSingleGlobGroup(globalIgnores: Array<GlobalIgnore>) {
+  return mergeGlobGroups(globalIgnores.map(({ ignores }) => ({ files: [], ignores })))
```
illright, line 6: issue: I'm kind of concerned with this operation that adds `files: []` to an object from a semantics perspective. The way the config file is defined, you can do either:
- specify just `files`, which will only include the files you specified
- specify both `files` and `ignores`, which will include the files you specified except for the ones that are covered by ignores
- specify only `ignores`, which will include all the files except for the ones that are covered by ignores

Therefore, files and ignores can be optional, and we should support that on the exclusion operator level. `{ files: [], ignores: ['*.js'] }` is not the same as `{ ignores: ['*.js'] }` semantically. The first one selects nothing, the second one selects everything except for top-level JS files.

suggestion: let's remove this mapping and adjust the exclusion operator to differentiate between these two cases

illright (changes requested):

illright:
I'm not done reviewing yet, thought I'd drop a few early comments

packages/steiger/src/features/calculate-diagnostic-severities/calculate-final-severity.ts
```diff
@@ -0,0 +1,22 @@
...
+export default function calculateFinalSeverities(
+  vfs: Folder,
+  ruleName: string,
+  paths: Array<string>,
+): Array<Exclude<Severity, 'off'>> {
+  const globGroups = getGlobsForRule(ruleName)
+
+  if (!globGroups) {
+    return []
+  }
```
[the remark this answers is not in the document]
daniilsapa: Yes, makes sense, I'll fix that

packages/steiger/src/features/remove-global-ignores-from-vfs/remove-global-ignores-from-vfs.ts
daniilsapa: This was a kind of foolproof mechanism on my part (to sort out situations where users think that an empty array simply selects everything (when in files) or does nothing (when in ignores)), allowing empty files/ignores arrays in these objects to be treated as if they weren't there at all. Because I didn't see any useful applications of such arrays when they were empty. There is literally no case where a user wants to use them intentionally. But, ok, I can make it more strict if you think it should be. Or alternatively, I can handle cases when they are empty more gracefully (without the need to add empty file arrays before glob exclusion). I think we can allow empty arrays in raw configs but remove them during the config parsing stage, so all other modules are not infected with this knowledge. What do you think?

illright: I would rather keep the logic predictable and do any fool-proofing on the config validation stage by printing warnings about potential errors. Let's make it more strict for now and we can always improve on the UX later if it turns out to be a problem.

> I think we can allow empty arrays in raw configs but remove them during the config parsing stage

That would be quite unpredictable, so I don't think we should go with this approach

packages/steiger/src/features/calculate-diagnostic-severities/calculate-final-severity.spec.ts
```diff
@@ -0,0 +1,61 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { joinFromRoot, parseIntoFsdRoot } from '../../_lib/prepare-test'
+import calculateFinalSeverities from './calculate-final-severity'
+import { GlobGroupWithSeverity } from '../../models/config'
+
+vi.mock('../../models/config', async () => {
+  const ruleToGlobs: Record<string, Array<GlobGroupWithSeverity>> = {
+    rule1: [
+      { files: [], ignores: [], severity: 'warn' },
```
illright, line 10: And related to that discussion, I think we should remove empty arrays here
```suggestion
      { severity: 'warn' },
```

packages/steiger/src/features/run-rule/prepare-vfs-for-rule-run.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        files: ['/src/pages/**'],
+        ignores: [],
+        severity: 'off',
+      },
+      {
+        files: ['**/ui/**'],
+        ignores: [],
+        severity: 'warn',
+      },
+    ] as Array<GlobGroupWithSeverity>
```
illright, line 80: And here as well:
```suggestion
      {
        severity: 'error',
      },
      {
        files: ['/src/pages/**'],
        severity: 'off',
      },
      {
        files: ['**/ui/**'],
        severity: 'warn',
      },
    ] as Array<GlobGroupWithSeverity>
```

packages/steiger/src/models/config/transform-globs.ts
```diff
@@ -0,0 +1,50 @@
...
+
+function convertRelativeGlobsToAbsolute(rootPath: string, globs: Array<string>) {
+  function composeAbsolutePath(root: string, glob: string) {
+    // Remove '/'. The root has platform-specific separators
+    const segmentsOfRoot = root.slice(1).split(sep)
+
+    return `/${posix.join(...segmentsOfRoot, glob)}`
+  }
+
+  return globs.map((glob) => (glob.startsWith('.') ? composeAbsolutePath(rootPath, glob) : glob))
```
illright, line 14: issue (non-blocking): I think we should be checking if the glob starts with `/` to tell if it's absolute or not. Otherwise we will not convert a glob like `"src/entities/**"`, which is a valid relative path, just the `./` is omitted.

suggestion: let's check if the path is absolute with [`path.isAbsolutePath`](https://nodejs.org/api/path.html#pathisabsolutepath) and add this case to the tests

packages/steiger/src/shared/globs/apply-exclusion.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+        files: ['/src/pages/account/**'],
+        ignores: [],
+      }),
+    ]
+
+    expect(applyExclusion(vfs, globs)).toEqual(expectedVfs)
+  })
+
+  it('should correctly apply exclusions for a special case', () => {
+    const globs = [not({ files: [], ignores: [] }), { files: ['/src/shared/ui/Button.ts'], ignores: [] }]
```
illright, line 175: question: what makes this case special?

daniilsapa: Yes, it looks like I should have renamed this test case, but I forgot to do so. Thanks for catching that!

packages/steiger/src/features/remove-global-ignores-from-vfs/remove-global-ignores-from-vfs.ts
daniilsapa: Got it, I don't mind and have made the necessary fixes

packages/steiger/src/shared/globs/merge-glob-groups.ts
```diff
@@ -0,0 +1,17 @@
...
+      globGroup.files = [...(acc.files || []), ...files]
+    }
+
+    if (ignores) {
+      globGroup.ignores = [...(acc.ignores || []), ...ignores]
+    }
+
+    return globGroup
+  })
+}
```
illright, line 17: issue: this function is throwing away globs if `files` or `ignores` is not present in the last element:
```ts
mergeGlobGroups({ files: ['*'], ignores: ['*'] }, {}) // returns {}
```

illright (changes requested):
One little bug to fix

packages/steiger/src/shared/globs/merge-glob-groups.ts
daniilsapa: Oh, yeah, I forgot to spread the previous result. Thanks for catching that!

illright (commented):
Okay, I think this is almost good to go! Could you please run https://knip.dev to get rid of some unused code and exports? The solution itself looks good :)

PR #96, opened 2024-09-02
New config structure docs

### LINKED ISSUE #89: Add information about the new config structure to README.md
- Description of the new config format
- How to use the new config
- A short guide to migration to the new config

The migration guide should contain commands to perform migration for the major package managers:
- npm
- yarn
- pnpm

daniilsapa:
Resolves #89

Hi @illright 

I added several new sections to README to describe the new structure, its concepts, and how to work with it. Also, I added a migration guide in a separate file, as you requested.

Let me know if you have any suggestions/corrections. Thank you in advance!


MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,47 @@
...
+
+**! Don't forget to check the changes after the migration and bring them to the code style adopted in your project**
+
+**pnpm**:
+
+```shell
+pnpx jscodeshift -t https://raw.githubusercontent.com/feature-sliced/steiger/codemod/packages/steiger/migrations/convert-config-to-flat.js steiger.config.js
+```
+
+_yarn_:
```
illright, line 37: nitpick (non-blocking): there's a formatting inconsistency here, other times the package managers are in `** **`, but here and after they are in `_ _`

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,47 @@
...
+yarn add @feature-sliced/steiger-plugin
+```
+
+**npm**:
+
+```shell
+npm i @feature-sliced/steiger-plugin
+```
+
+**Step 3**: The final step. You need to bring your `steiger.config.js` file to the new configuration format. You can do that automatically by running one of the following commands.
```
illright, line 27: suggestion: since the migration script is imperfect (and the command is also long), I'd like to include a bit about what it actually does. For example, show a "before" and "after". And also, if people have a `steiger.config.ts` instead of `.js`, we should let them know that they need to modify this command.

As for the "before" and "after", we can do it like this in Markdown:

    <table><thead><tr>
    <th>Before</th>
    <th>After</th>
    </tr></thead><tbody><tr><td>
    
    ```ts
    // steiger.config.ts
    ```
    
    </td><td>
    
    ```ts
    // steiger.config.ts
    ```
    
    </td></tr></tbody></table>

The empty lines are there to prevent HTML from messing up Markdown. It will produce this:

<table><thead><tr>
<th>Before</th>
<th>After</th>
</tr></thead><tbody><tr><td>

```ts
// steiger.config.ts
```

</td><td>

```ts
// steiger.config.ts
```

</td></tr></tbody></table>

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,47 @@
...
+## From 0.4.0 to 0.5.0
+
+**Step 1**: First of all you need to make sure you upgraded Steiger package to version 0.5.0 or higher.
+
+**Step 2**: You need to install `@feature-sliced/steiger-plugin` package (that contains all FSD rules to run checks in your project). Run one of the following commands based on the package manager you use.
+
+**pnpm**:
+
+```shell
+pnpm add @feature-sliced/steiger-plugin
```
illright, line 12: suggestion: I think this package should be installed to dev dependencies

illright (changes requested):
Thanks for the detailed write-up! 

One major concern I have here is the size of the README. Ideally, I'd like to keep the README short, and have a separate resource for plugin authors and people who want to quickly write a rule or two. In the near future, it will probably amount to a documentation site. That's why I don't think we should go into details about concepts like GlobArray, Config Object, etc. These are implementation details and terms rather than public facing components, in my opinion.

So maybe we can just keep the README to one example that includes `fsd.configs.recommended`, a global ignore, and then a change of severity for a certain glob?

As for the docs about writing plugins, I think we should hold on with this for now. Since we still have some technical and design issues to iron out (mostly on my side, with the toolkit stuff), I don't think we're ready to write docs just yet.

MIGRATION_GUIDE.md
daniilsapa: 100%, missed that thing

MIGRATION_GUIDE.md
daniilsapa: Yeah, indeed. Thanks for catching that

daniilsapa:
I fixed all the issues. I decided not to delete the examples completely and moved them to a separate page.

README.md
```diff
@@ -31,23 +34,59 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
+import defineConfig from 'steiger'
+
+export default defineConfig([
+  ...fsd.configs.recommended,
+  {
+    // ignore all mock files for all rules
+    ignores: ['**/__mocks__/**'],
+  },
+  {
+    files: ['**/shared/**'],
```
illright, line 66: suggestion (non-blocking): ideally I think we should encourage people to provide precise paths to layers, otherwise if there's a folder called "shared" somewhere deep within, it will also be targeted
```suggestion
    files: ['./src/shared/**'],
```

illright (changes requested):

illright:
Sorry for not finding these comments before, the fixes for the previous comments are good, thanks!

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+</td></tr></tbody></table>
+
+**! Don't forget to check the changes after the migration and bring them to the code style adopted in your project**
+
+**pnpm**:
+
+```shell
+pnpx jscodeshift -t https://raw.githubusercontent.com/feature-sliced/steiger/codemod/packages/steiger/migrations/convert-config-to-flat.js steiger.config.js
```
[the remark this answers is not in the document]
daniilsapa: Oh, yes, thanks for catching that!

README.md
```diff
@@ -31,23 +34,59 @@ To run in watch mode, add `-w`/`--watch` to the command:
 steiger ./src --watch
 ```
 
-# Configuration
+## Configuration
 
 Steiger is zero-config! If you don't want to disable certain rules, you can safely skip this section.
 
 Steiger is configurable via `cosmiconfig`. That means that you can create a `steiger.config.ts` or `steiger.config.js` file in the root of your project to configure the rules. Import `{ defineConfig } from "steiger"` to get autocompletion.
```
illright, line 41: issue: this section is no longer accurate with the new changes

suggestion: let's leave the type safety for a future PR
```suggestion
Steiger is configurable via `cosmiconfig`. That means that you can create a `steiger.config.ts` or `steiger.config.js` file in the root of your project to configure the rules.
```

README.md
```diff
@@ -31,23 +34,59 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
 
-```ts
-import { defineConfig } from 'steiger'
+The config file shape is highly inspired by ESLint's config file, so if you have configured ESLint before, you'll find it easy to configure Steiger.
+
+### Example
+
+Here are some rules on how configuration is processed:
+
+- Config objects are processed from top to bottom, so if there are multiple config object that match the same file for the same rule, the last one will be applied.
```
illright, line 49: issue: I think it might be slightly unclear what "config objects" are, they might be confused with the entire file.

suggestion:
```suggestion
- Objects in the config array are processed in order, so later objects can override the objects before them.
```

README.md
```diff
@@ -31,23 +34,59 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
+- You can set options for a rule once. When set, options are applied for the entire file system that is covered by Steiger.
```
illright, line 50: issue: I think we also need to clarify here that it's not just that you can set it once, you should set it once

suggestion:
```suggestion
- If a rule has configurable options, then you can set it once in any of the objects in the array and these options will be applied to the entire codebase, regardless of `files` and `excludes`.
```

illright (changes requested):
Actually, when reading the README in rendered Markdown, I noticed something else. One of them is a thing that is straight-up no longer true, others are simply clarifications

daniilsapa:
Yes, your suggestions were completely reasonable, so I committed them

daniilsapa:
Also, I thought about merging this PR last, because:
- While we are buttoning up the new config there can be slight changes to the documentation (e.g. what we decided about GlobalIgnores today)
- Let's not confuse people with documentation for features that haven't been released yet.

illright:
Makes sense. Let's wait with this then

daniilsapa:
@illright 
FYI I'm going to update this PR with some information on how our glob matching logic works. Let me know if we need to add anything else.

README.md
```diff
@@ -7,19 +7,22 @@ Universal file structure and project architecture linter.
 > [!NOTE]
 > The project is in beta and in active development. Some APIs may change.
 
-# Features
+> [!NOTE]
+> Version 0.5.0 introduces a new config file format. Please refer to the [Configuration](#configuration) section for more information and migration guide.
```
illright, line 11: issue: the migration guide is quite far down that section

suggestion: let's just put the link here:
```suggestion
> Version 0.5.0 introduced a new config file format. We have a codemod to automatically update your config, see the [migration guide](./MIGRATION_GUIDE.md).
```

README.md
```diff
@@ -31,23 +34,70 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
 
 Steiger is configurable via `cosmiconfig`. That means that you can create a `steiger.config.ts` or `steiger.config.js` file in the root of your project to configure the rules. Import `{ defineConfig } from "steiger"` to get autocompletion.
 
-```ts
-import { defineConfig } from 'steiger'
+The config file shape is highly inspired by ESLint's config file, so if you have configured ESLint before, you'll find it easy to configure Steiger.
+
+### Example
+
+Here are some rules on how configuration is processed:
```
illright, line 47: issue: when I see the header "Example", I expect to see a simple example right away, but in this version there's a big blob of text to go through first :D

suggestion: let's put a small example first, like this one:
```js
// ./steiger.config.js
import { defineConfig } from 'steiger'

export default defineConfig([
  ...fsd.configs.recommended,
  {
    // disable the `public-api` rule for files in the Shared layer
    files: ['./src/shared/**'],
    rules: {
      'fsd/public-api': 'off',
    },
  }
])
```

README.md
```diff
@@ -31,23 +34,70 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
 
-export default defineConfig({
-  rules: {
-    'public-api': 'off',
+- Config objects are processed from top to bottom, so if there are multiple config object that match the same file for the same rule, the last one will be applied.
+- GlobalIgnore objects `{ ignores: ['**/__mocks__/**'] }` are applied to all rules, they are processed first and permanently remove files from the linter's field of view, so you can't reassign severity to them in other later objects.
+- You can set options for a rule once. When set, options are applied for the entire file system that is covered by Steiger.
+
+Note that this line `...fsd.configs.recommended,` just takes the plugin and the recommended rules configuration (all enabled with "error" severity by default) and puts it into the config array.
```
illright, line 53: suggestion: let's not try to document every single aspect of behaviour. I think this makes instructions too big, which means that people will be less likely to read them.

For instance:
- "Config objects are processed from top to bottom" — I think this is something that they can discover by trying, and most people are probably not even going to encounter that
- "GlobalIgnore objects" — I think this goes a bit too deep into the internal terminology like "GlobalIgnore", "assign severity", "linter's field of view", I don't think that's super relevant. I suggest simplifying it
- "You can set options for a rule once." — at present, none of the rules even accept options, I think, so this is very much irrelevant :D
- About the `fsd.configs.recommended`, I don't think we need to tell people how it works exactly, all they need to know is that it will set up the linter the way we recommend, and they can tweak it further if they want.

I don't even think it should be a list, let's just throw a tip like this after that simple example:
```suggestion
> [!TIP]
> If you want Steiger to ignore certain files, add an object like this to the config array: 
> ```js
> defineConfig([
>   /* … */, 
>   { ignores: ['**/__mocks__/**'] }
> ])
> ```
```

README.md
```diff
@@ -31,23 +34,70 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
+
+```javascript
+// ./steiger.config.ts
+import fsd from '@feature-sliced/steiger-plugin'
+import defineConfig from 'steiger'
```
illright, line 58: suggestion: let's put this example under a spoiler so that it's there if you want to go further, but doesn't clutter the README for the people who want to speed-run the README:

```
<details>
  <summary>Comprehensive showcase of the config file syntax</summary>
  
```js
...
```

</details>
```

Comment (illright) [reply README.md:58]: also, issue: the example shows `import defineConfig from 'steiger'`, it should be `import { defineConfig } from 'steiger'`

File: README.md [hunk 13]
```diff
@@ -31,23 +34,70 @@ To run in watch mode, add `-w`/`--watch` to the command:
...
+
+All globs are matched only against files, folder severities are computed based on the files inside them. The formula is simple: the folder severity is the highest severity of files inside it (from highest to lowest: error, warn, off).
+
+**Glob examples**:
+
+- `./src/shared/**` - matches all files in the `shared` folder and its subfolders
+- `./src/shared/*` - matches all files that are direct children of the `shared` folder
+- `./src/shared` - based on the fact that globs are matched against files, this one matches only `shared` file (without an extension) inside the `src` folder
+- `**/__mocks__/**` - matches all files in all `__mocks__` folders throughout the project
+- `**/*.{test,spec}.{ts,tsx}` - matches all test files in the project
```
illright, line 94: issue: again, I believe this is a more technical description, and I forecast that most people will just disable entire folders, so people are unlikely to run up against a case where one file made the whole folder be `error`.

suggestion: let's just get rid of this entire section

EXAMPLES.md
illright, line 1: suggestion: let's rename this file to `CONFIG_EXAMPLES.md` and link it in the README, under that spoiler with a comprehensive showcase

EXAMPLES.md
```diff
@@ -0,0 +1,160 @@
...
+      ignores: ['./src/shared/__mocks__/**'],
+    },
+  ])
+  ```
+
+Parts of the config object:
+
+- Severity - can be one of the following: "off", "warn", "error"
+- GlobArray - string array with glob patterns to match files and folders in your project.
+- RuleOptions - an object that contains specific rule options and can be passed to the rule to configure its behavior.
```
illright, line 47: suggestion: honestly, I'd love if we just deleted this entire chunk of text. The file is called examples, people learn wonderfully from examples, let's just throw a bunch of diverse examples at them and let them figure it out

and if they don't, we'll just revise the examples instead of giving them the technical readout

EXAMPLES.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Examples
+
+Here are some rules on how configuration is processed:
+
+- Config objects are processed from top to bottom, so if there are multiple config object that match the same file for the same rule, the last one will be applied.
+- You can set options for a rule once. When set, options are applied for the entire file system that is covered by Steiger.
+
+Note that this line `...fsd.configs.recommended,` just takes the plugin and the recommended rules configuration (all enabled with "error" severity by default) and puts it into the config array.
```
illright, line 56: oh, and this as well xD

EXAMPLES.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Example 1. Default case
+
+```javascript
+// ./steiger.config.ts
+import fsd from '@feature-sliced/steiger-plugin'
+import defineConfig from 'steiger'
```
illright, line 63: suggestion: also we should update all the examples to import `{ defineConfig }`

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+      'fsd/ambiguous-slice-names': 'off',
+      'fsd/no-processes': 'off',
+    },
+  },
+])
+```
+
+</td></tr></tbody></table>
+
+**! Don't forget to check the changes after the migration and bring them to the code style adopted in your project**
```
illright, line 70: nitpick (non-blocking): I find it a bit janky when a regular exclamation mark is used in the beginning of a sentence, especially when there's no punctuation at the end. Besides, this point is not that important that we need to scream about it in bold :)

suggestion: let's use GitHub's Markdown capabilities to our advantage:

```suggestion
> [!NOTE]
> The codemod will probably get the formatting wrong, so don't forget to check the file yourself after migration.
```

illright (changes requested):
Phew, sorry for this big load of corrections, it's just that documentation is something that I am particularly passionate and opinionated about. Let me know if you disagree with anything

README.md
daniilsapa: Yes, makes sense

README.md
daniilsapa: No problem

README.md
daniilsapa: I agree, most of this section is either irrelevant or too complex

README.md
daniilsapa: That sounds completely reasonable to me

EXAMPLES.md
daniilsapa: Okay, I'll stash this for now somewhere. It might come in handy a bit when it's time to create a document site

README.md
daniilsapa: Decided to move it to CONFIG_EXAMPLES.md so far. Let me know if you think it should not be even there

daniilsapa:
You don't need to be sorry for these corrections. They are completely reasonable and improve the documentation. Thanks for reviewing!

illright:
There's still some instances of `import defineConfig`. Could you search the project for this and replace it with `import { defineConfig }`?

README.md
illright: I personally think it shouldn't be even there, but I also don't mind it being there. As long as it's not up front in the README, I'm fine with it

illright (approved):
Thanks! I took the liberty to introduce some minor corrections to the examples myself, because I'm eager to merge this :) Let me know if they look alright to you, and then we can merge

daniilsapa:
Thanks for catching that! Yes, I forgot to update some examples based on the new glob-matching approach

Issue #77, opened 2024-08-04
Implement `no-ui-in-app`


illright:
It's unusual to have the `ui` segment on the App layer, because it's usually just the layer for putting everything together and connecting providers (note: providers are not UI)

PR #101, opened 2024-09-17
Add no-ui-in-app rule

Espate:
Resolves [#77](https://github.com/feature-sliced/steiger/issues/77)


illright:
Hi, thanks for taking a hand at it! I looked over the PR, looks solid, great job. One thing I'd like to add to it is to mention in the rationale of the rule that a valid use case for having a UI segment in App is for applications that only have a single page. They are rare, but they deserve a mention. 

Also I noticed a couple of grammatical errors in the rationale, could you please run it through a spell checker like https://languagetool.org/?

Espate:
> Hi, thanks for taking a hand at it! I looked over the PR, looks solid, great job. One thing I'd like to add to it is to mention in the rationale of the rule that a valid use case for having a UI segment in App is for applications that only have a single page. They are rare, but they deserve a mention.

Hi, will update this part.

> Also I noticed a couple of grammatical errors in the rationale, could you please run it through a spell checker like https://languagetool.org/?

Could you please point me out where exactly the errors are? I used the tool you referred to, and it didn't reveal any issues, at least in the free plan.

Screenshot 2024-09-18 at 08 45 13

packages/steiger-plugin-fsd/src/no-ui-in-app/README.md
```diff
@@ -0,0 +1,39 @@
...
+📂 app
+  📂 providers
+    📄 index.ts
+  📂 ui  // ❌
+    📄 index.ts
+```
+
+## Rationale
+
+This layer is an entry point of the application, it handles global logic such as routing initialization, dependency injection, different providers, stores, etc. As for <code>ui</code> segment, it belongs to view part of the application and typically defined within pages, features, entities or other underlying layers.
```
illright, line 39: Indeed, LanguageTool didn't really help here, sorry for not checking beforehand. I would want to add a couple of articles and reword the part about layers a bit:

```suggestion
This layer is an entry point of the application, it handles global logic such as routing initialization, dependency injection, different providers, stores, etc. As for the <code>ui</code> segment, it belongs to the "view" part of the application and is typically present in the layers below, for example, pages, features, entities.
```

Espate: Updated, please review it again.

illright (approved):
I played around with the rationale a bit more, let me know if that sounds okay. If so, then it's ready to merge! Thanks again :)

Espate:
> I played around with the rationale a bit more, let me know if that sounds okay. If so, then it's ready to merge! Thanks again :)

Sure, it works for me too. Thanks!

PR #102, opened 2024-09-19
Add typo-in-layer-name rule

### LINKED ISSUE #93: Implement `typo-in-layer-name`
If there's a folder among the top-level folders called `widgents`, for example, this rule should warn that it was probably `widgets` that was intended

Espate:
Resolves #93 

I checked a few packages which have built-in spellcheck feature, but they don't fit the needs due to:
- require additional dictionary, e.g. in `.aff` or `.dic` formats ([nodehun](https://github.com/Wulf/nodehun), [nspell](https://github.com/wooorm/nspell), [typojs](https://github.com/cfinke/Typo.js))
- archived repo (maybe not crucial, but still) ([node-spellchecker](https://github.com/atom/node-spellchecker))
- have a huge size ([natural](https://github.com/NaturalNode/natural), ~7MB)

So, I decided to use lightweight & fast package called [fastest-levenshtein](https://github.com/ka-weihe/fastest-levenshtein) which implements [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) algorithm.

### DIFF
--- a/.changeset/unlucky-coins-accept.md
+++ b/.changeset/unlucky-coins-accept.md
@@ -1,0 +1,6 @@
+---
+'@feature-sliced/steiger-plugin': minor
+'steiger': minor
+---
+
+Add typo-in-layer-name rule
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Currently, Steiger is not extendable with more rules, though that will change in
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/repetitive-naming/README.md"><code>repetitive-naming</code></a></td> <td>Ensure that all entities are named consistently in terms of pluralization.</td> </tr>
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/segments-by-purpose/README.md"><code>segments-by-purpose</code></a></td> <td>Discourage the use of segment names that group code by its essence, and instead encourage grouping by purpose</td> </tr>
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/shared-lib-grouping/README.md"><code>shared-lib-grouping</code></a></td> <td>Forbid having too many ungrouped modules in <code>shared/lib</code>.</td> </tr>
+  <tr> <td><a href="./packages/steiger-plugin-fsd/src/typo-in-layer-name/README.md"><code>typo-in-layer-name</code></a></td> <td>Ensure that all layers are named without any typos.</td> </tr>
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/no-processes/README.md"><code>no-processes</code></a></td> <td>Discourage the use of the deprecated Processes layer.</td> </tr>
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/import-locality/README.md"><code>import-locality</code></a></td> <td>Require that imports from the same slice be relative and imports from one slice to another be absolute.</td> </tr>
 </tbody>
--- a/packages/steiger-plugin-fsd/package.json
+++ b/packages/steiger-plugin-fsd/package.json
@@ -31,6 +31,7 @@
   ],
   "dependencies": {
     "@feature-sliced/filesystem": "^2.2.5",
+    "fastest-levenshtein": "^1.0.16",
     "lodash-es": "^4.17.21",
     "pluralize": "^8.0.0",
     "precinct": "^12.1.1",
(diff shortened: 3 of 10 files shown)


illright (changes requested):
Thanks for a thorough research of potential dependencies and a quick and solid solution! As usual, I have a suggestion regarding the rationale, but this time I also think we should make the rule more sophisticated.

Right now, it doesn't set an upper boundary on the levenshtein distance. In practice, that means that if I have a folder called `templates` in the root, it will be treated as a typo from "pages" (distance: 5), even though it's clearly not a typo, and has nothing to do with pages. Moreover, we should consider a case when the layer that we think was meant is actually present. So if there is `src/pages` and `src/templates`, the linter saying that `src/templates` is meant to be `src/pages` is even more ridiculous, as there already is a `pages`. Another thing to consider here is if a name of a folder has several words like `license-files`.

Ideally, we would bring in an actual typo detector, like the `nspell` project you mentioned, along with `dictionary-en`, which would be 500KB, a bit too much. I think we can still get by without it, by doing a matrix of levenshtein distances (LDs) between all available folders and all available layers, then picking out exactly one pair that is closest by LD and removing it from the search, so that no other pair can be formed with either of these. And we should probably consider 3 as the largest LD for a typo, so if it's more than that, then it's probably not a typo, but instead a different intention.

Espate:
> Right now, it doesn't set an upper boundary on the levenshtein distance. In practice, that means that if I have a folder called `templates` in the root, it will be treated as a typo from "pages" (distance: 5), even though it's clearly not a typo, and has nothing to do with pages. Moreover, we should consider a case when the layer that we think was meant is actually present. So if there is `src/pages` and `src/templates`, the linter saying that `src/templates` is meant to be `src/pages` is even more ridiculous, as there already is a `pages`. Another thing to consider here is if a name of a folder has several words like `license-files`.
 
Good point, I thought the methodology forbids any custom layers, so it's not needed to take into account `templates` or `license-files` as valid inputs.

> I think we can still get by without it, by doing a matrix of levenshtein distances (LDs) between all available folders and all available layers, then picking out exactly one pair that is closest by LD and removing it from the search, so that no other pair can be formed with either of these. And we should probably consider 3 as the largest LD for a typo, so if it's more than that, then it's probably not a typo, but instead a different intention.

Let's talk about this a bit. Am I right that your proposal is to build a matrix like this?

|           | app | page | shraed | wigdets | fiatures |
|-----------|-----|------|--------|---------|----------|
| app       |  0  |  3   |   5    |    7    |    7     |
| pages     |  4  |  1   |   5    |    4    |    5     |
| shared    |  5  |  4   |   2    |    6    |    5     |
| widgets   |  7  |  5   |   6    |    2    |    6     |
| features  |  7  |  6   |   7    |    7    |    1     |

Then fix & exclude `app` (0), `page` | `fiatures` (1), `shraed` (2), etc.

If so, I guess it might be overhead a bit due to it's uncommon to have such amount of typos (1 is ok, maybe 2, but more - doubtful). Besides, if we set upper bound of Levenshtein distance as 3, there will not be a situation where there is a typo in the name of one layer, but the rule suggests another layer, because Levenshtein distance will be more than 3 and the rule will not complain at all.

illright:
> I thought the methodology forbids any custom layers

It does, but some frameworks like Nuxt insist on having some extra folders, so we need to prepare for that

> Let's talk about this a bit. Am I right that your proposal is to build a matrix like this?

Yes, exactly

> If so, I guess it might be overhead a bit due to it's uncommon to have such amount of typos (1 is ok, maybe 2, but more - doubtful). Besides, if we set upper bound of Levenshtein distance as 3, there will not be a situation where there is a typo in the name of one layer, but the rule suggests another layer, because Levenshtein distance will be more than 3 and the rule will not complain at all.

I didn't fully understand this bit, could you elaborate please? As for the overhead, I think it's manageable, because it's 6 × <number of folders>, it's okay both in terms of memory and in terms of computing the LD (the library does it pretty fast)

illright:
Ah, wait, I understand it now. Well, either way, we have to be able to tell that "shentities" is more probably entities than it is shared. I can come up with a more practical example later, but I think it's necessary to do the matrix regardless of where we set the upper bound

Espate:
> Ah, wait, I understand it now. Well, either way, we have to be able to tell that "shentities" is more probably entities than it is shared.

That sounds reasonable. Agreed, in that case matrix will provide better suggestions.

> I can come up with a more practical example later, but I think it's necessary to do the matrix regardless of where we set the upper bound

We will be able to add it in specs, so it would be really useful.

illright:
I found all the words that might technically somehow be in a project with distances 1 or 2 from the layer names:
- shared: `shard, hard, sealed, saved, shapes, staged`
- entities: `entries, fixtures, futures, gestures`
- pages: `cases, images, packs, panels, paths, places, saves, pip, zip`
- app: `amp, apt`

illright (approved):
Great stuff!

PR #104, opened 2024-10-10
Upgrade dependencies

illright:
There are some vulnerabilities that Dependabot complains about, but it's a good idea to update anyway


daniilsapa (approved):
Reviewed, it's good to go!

PR #105, opened 2024-10-11
Throw an error for old config

### LINKED ISSUE #90: Throw an exception when the old config structure is detected
The error message should contain instructions on how to fix it.

daniilsapa:
Resolves #90

I added error messages for invalid config shapes. Also, I refactored config validation a bit.

It looks like we need to improve the pretty-printer a bit so we can delegate printing not just diagnostics but also other service error messages and streamline that process too. Hence, the service messages have interfaces and are collected in a predictable and orderly manner. I decided to not overengineer the solution yet because it's a separate story that needs a preliminary discussion.

### DIFF
--- a/packages/steiger/src/models/config/validate-config.ts
+++ b/packages/steiger/src/models/config/validate-config.ts
@@ -1,8 +1,16 @@
 import z from 'zod'
 
+import { BaseRuleOptions, Config, Plugin, Severity } from '@steiger/types'
+
 import { getOptions, isConfigObject, isPlugin } from './raw-config'
 import { isEqual } from '../../shared/objects'
-import { BaseRuleOptions, Config, Plugin, Severity } from '@steiger/types'
+
+const OLD_CONFIG_ERROR_MESSAGE =
+  'Old configuration structure noticed. We are evolving!\nPlease follow these easy 5-minute instructions to migrate to the new one:\nhttps://github.com/feature-sliced/steiger/blob/master/MIGRATION_GUIDE.md'
+const WRONG_CONFIG_SHAPE_ERROR_MESSAGE =
+  'The provided config has a wrong shape.\nPlease check your configuration file and follow the documentation to fix it:\nhttps://github.com/feature-sliced/steiger?tab=readme-ov-file#configuration'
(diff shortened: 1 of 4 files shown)


packages/steiger/src/models/config/validate-config.ts
```diff
@@ -1,8 +1,16 @@
...
+import { BaseRuleOptions, Config, Plugin, Severity } from '@steiger/types'
+
 import { getOptions, isConfigObject, isPlugin } from './raw-config'
 import { isEqual } from '../../shared/objects'
-import { BaseRuleOptions, Config, Plugin, Severity } from '@steiger/types'
+
+const OLD_CONFIG_ERROR_MESSAGE =
+  'Old configuration structure noticed. We are evolving!\nPlease follow these easy 5-minute instructions to migrate to the new one:\nhttps://github.com/feature-sliced/steiger/blob/master/MIGRATION_GUIDE.md'
+const WRONG_CONFIG_SHAPE_ERROR_MESSAGE =
+  'The provided config has a wrong shape.\nPlease check your configuration file and follow the documentation to fix it:\nhttps://github.com/feature-sliced/steiger?tab=readme-ov-file#configuration'
```
illright, line 11: suggestion: reworded the messages slightly. Some changes are because it sounds more natural to me that way, others, like the migration guide, I felt we should change because we don't get to decide if the guide is easy, or that it takes 5 minutes :)

```suggestion
const OLD_CONFIG_ERROR_MESSAGE =
  'Old configuration format detected. We are evolving!\nPlease follow this little guide to migrate to the new one:\nhttps://github.com/feature-sliced/steiger/blob/master/MIGRATION_GUIDE.md'
const WRONG_CONFIG_SHAPE_ERROR_MESSAGE =
  'The config should be an Array, but the provided config is not.\nHere is a link to the documentation that might help to fix it:\nhttps://github.com/feature-sliced/steiger?tab=readme-ov-file#configuration'
```

illright (changes requested):
Nice work! I'd prefer if we changes the messages a bit, but other than that, the code is good

illright (approved):
Great!

PR #108, opened 2024-10-19
Throw errors on non-existent paths

### LINKED ISSUE #58: Throw an exception if a non-existent folder is provided as an FSD root.
Right now it just says that no problems were found. But it seems the tool should exit with an error in this case, so the user can understand that they misconfigured the tool early.

Screenshot 2024-07-16 at 19 59 55

illright:
Closes #58

### DIFF
--- a/.changeset/mean-buses-tan.md
+++ b/.changeset/mean-buses-tan.md
@@ -1,0 +1,5 @@
+---
+'steiger': patch
+---
+
+Fix the non-existent paths showing up as a success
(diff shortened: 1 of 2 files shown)


daniilsapa:
Great updates! I don't have any suggestions or corrections. Just one question.

Why did you use these specific exit codes, 101 and 102? I did some googling about this, but I couldn't find any conventions about these exact codes for NodeJS or CLI tools in general. Are they something we create our own convention about? If so, should we keep track of such codes? Just create a document with an internal registry of such codes, so we know which codes are already taken and for what purpose

illright:
I just used some random codes that weren't used before. I don't think we need to keep track of them separately, because they're only in cli.ts for now, easy enough to remember.

If you don't have any objections, could you please approve? :) otherwise GitHub won't let me merge

daniilsapa (approved):

Issue #109, opened 2024-10-19
Implement the JSON reporter


illright:
Right now, the CLI prints the diagnostics in a human-readable format, which is nice, but not scriptable. We should add a new option `--reporter` to the CLI (no need for a shorthand) that accepts options `pretty` (default) or `json`. When `--reporter json`, the output should be pretty-printed JSON, no color. Basically, just dump the whole diagnostics array as it is.

Benjamin-Frost:
Hi @illright, I'm interested in contributing to this project for Hacktoberfest and would love to work on this issue. Could I take a shot at it?

illright:
Hi @Benjamin-Frost, that's great! Yes, please go ahead, and feel free to ask any questions here if you get stuck :)

PR #114, opened 2024-10-26
Add JSON Reporter

Benjamin-Frost:
Closes: #109

This PR adds a `reporter` option to the CLI, allowing output in either pretty-print (default) or JSON.

Let me know if this is what you had in mind! I'm happy to make changes, if you have any suggestions :)

### DIFF
--- a/packages/steiger/src/app.ts
+++ b/packages/steiger/src/app.ts
@@ -69,3 +69,5 @@ export const linter = {
 export function defineConfig<Rules extends Array<Rule> = Array<Rule>>(config: Config<Rules>) {
   return config
 }
+
+export type Diagnostics = Awaited<ReturnType<typeof runRules>>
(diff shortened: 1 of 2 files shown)


packages/steiger/src/app.ts
```diff
@@ -69,3 +69,5 @@ export const linter = {
 export function defineConfig<Rules extends Array<Rule> = Array<Rule>>(config: Config<Rules>) {
   return config
 }
+
+export type Diagnostics = Awaited<ReturnType<typeof runRules>>
```
illright, line 73: suggestion: this type is essentially `Array<Diagnostic>`, where `Diagnostic` comes from `@steiger/types`. I'd suggest to refer to that type directly instead of getting the return type from `runRules`

illright (changes requested):
Thanks for solving this so quickly! Yes, that's exactly what I had in mind :) One little correction though, just to maintain a single source of truth regarding types

packages/steiger/src/app.ts
Benjamin-Frost: Ah, I didn't see that type – makes a lot of sense though. Thanks for the quick review!

illright (approved):
Awesome! Ready to merge

PR #115, opened 2024-10-27
Relax public API requirements on shared/lib and shared/ui

### LINKED ISSUE #76: Public API for shared/lib
The `no-public-api-sidestep` rule should consider the specifics of `shared/lib` — its public API should be one layer deeper, in `shared/lib/*/index`

illright:
Closes #76

### DIFF
--- a/packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.spec.ts
+++ b/packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.spec.ts
@@ -1,4 +1,4 @@
-import { expect, it, vi } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 
 import { joinFromRoot, parseIntoFolder as parseIntoFsdRoot } from '@steiger/toolkit'
 import noPublicApiSidestep from './index.js'
@@ -16,10 +16,16 @@ vi.mock('node:fs', async (importOriginal) => {
 
   return createFsMocks(
     {
+      '/shared/i18n/index.ts': '',
+      '/shared/i18n/translator.ts': '',
       '/shared/ui/styles.ts': '',
       '/shared/ui/Button.tsx': 'import styles from "./styles";',
       '/shared/ui/TextField.tsx': 'import styles from "./styles";',
       '/shared/ui/index.ts': '',
+      '/shared/lib/index.ts': '',
+      '/shared/lib/dates.ts': '',
+      '/shared/lib/i18n/index.ts': '',
+      '/shared/lib/i18n/translator.ts': '',
       '/entities/user/@x/product.ts': '',
       '/entities/user/ui/UserAvatar.tsx': 'import { Button } from "@/shared/ui"',
       '/entities/user/index.ts': '',
@@ -32,8 +38,13 @@ vi.mock('node:fs', async (importOriginal) => {
       '/pages/editor/ui/EditorPage.tsx': 'import { Button } from "@/shared/ui"; import { Editor } from "./Editor"',
       '/pages/editor/ui/Editor.tsx':
         'import { TextField } from "@/shared/ui"; import { ProductCard } from "@/entities/product/ui/ProductCard.tsx"',
-      '/pages/editor/ui/SubmitButton.tsx': 'import { Button } from "@/shared/ui/Button"',
+      '/pages/editor/ui/SubmitButton.tsx':
+        'import { Button } from "@/shared/ui/Button"; import { translator } from "@/shared/i18n/translator"',
       '/pages/editor/index.ts': '',
+      '/pages/settings/index.ts': '',
+      '/pages/settings/ui/SettingsPage.tsx':
+        'import { Button } from "@/shared/ui"; import { dates } from "@/shared/lib/dates"; import { translator } from "@/shared/lib/i18n"',
+      '/pages/settings/ui/Password.tsx': 'import { translator } from "@/shared/lib/i18n/translator";',
     },
     originalFs,
   )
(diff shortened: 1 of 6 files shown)


packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.ts
```diff
@@ -64,12 +65,30 @@ const noPublicApiSidestep = {
...
 
-          const index = getIndex(segment)
-          if (resolvedDependency !== index?.path) {
-            diagnostics.push({
-              message: `Forbidden sidestep of public API when importing from "${dependency}".`,
-              location: { path: sourceFile.file.path },
-            })
+          const segmentIndex = getIndex(segment)
+          if (segment.type === 'folder' && resolvedDependency !== segmentIndex?.path) {
+            if (dependencyLocation.layerName === 'shared' && ['ui', 'lib'].includes(dependencyLocation.segmentName)) {
```
daniilsapa, line 70: Non-blocking: this set of conditions is pretty difficult to read. I would create several variables that could make it possible to read it almost like plain English e.g. `isSharedLayer`, `isUIorLib` etc. What do you think

packages/steiger-plugin-fsd/src/public-api/index.ts
```diff
@@ -15,20 +15,38 @@ const publicApi = {
...
+                    path: join(segment.path, `index.js`),
+                    // TODO: Infer better content for this file
+                    content: '',
+                  },
+                ],
+                location: { path: segment.path },
+              })
+            } else if (segment.type === 'folder') {
+              for (const child of segment.children) {
+                if (child.type === 'folder' && getIndex(child) === undefined) {
```
daniilsapa, line 35: I saw the discussion on the issue. You mentioned that it should work either with an index file for the segment or with index files in the top-level folder inside the segment. Did you decide to go just with top-level folders, or am I wrong?

daniilsapa: Also, I didn't find a check for the shared level. Should it work for all ui/lib segments throughout the codebase?

packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.spec.ts
```diff
@@ -115,8 +129,57 @@ it('reports errors on a project with a public API sidestep on shared', async ()
...
+    const root = parseIntoFsdRoot(`
+      📂 shared
+        📂 ui
+          📄 styles.ts
+          📄 Button.tsx
+          📄 TextField.tsx
+          📄 index.ts
+        📂 lib
+          📄 dates.ts
+          📂 i18n
```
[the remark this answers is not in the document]
illright: No, that was actually intentional. I see both shared/i18n and shared/lib/i18n as valid ways of organizing i18n related code

packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.ts
illright: Yeah, I also thought that, but didn't know how to improve. I'll try your suggestion

packages/steiger-plugin-fsd/src/public-api/index.ts
illright: > Did you decide to go just with top-level folders, or am I wrong?

That part of the code is only entered when there is no segment-level index, so both are supported.

In the issue discussion, I also mentioned that it should be one or the other, not both, but I later realized that there is no point in enforcing this

As for checking for the Shared layer, the check is implicit. We only enter that code on unsliced layers (App and Shared) and the App layer case exits early

packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.spec.ts
```diff
@@ -16,10 +16,16 @@ vi.mock('node:fs', async (importOriginal) => {
...
   return createFsMocks(
     {
+      '/shared/i18n/index.ts': '',
+      '/shared/i18n/translator.ts': '',
       '/shared/ui/styles.ts': '',
       '/shared/ui/Button.tsx': 'import styles from "./styles";',
       '/shared/ui/TextField.tsx': 'import styles from "./styles";',
       '/shared/ui/index.ts': '',
+      '/shared/lib/index.ts': '',
+      '/shared/lib/dates.ts': '',
```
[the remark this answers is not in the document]
illright: We already have `'shared/lib/i18n/index.ts'`, if I understand you correctly

packages/steiger-plugin-fsd/src/no-public-api-sidestep/index.ts
illright: Decided to go with this in the end:
```ts
const { layerName, segmentName } = dependencyLocation
if (layerName === 'shared' && ['ui', 'lib'].includes(segmentName)) {}
```

daniilsapa: That works for me, thanks!

packages/steiger-plugin-fsd/src/public-api/index.ts
daniilsapa: Oh, got it

daniilsapa: And yes, it looks like enforcing one or the other way is unnecessary, I agree

daniilsapa:
Thank you for addressing the call-outs!

daniilsapa (approved):

PR #117, opened 2024-10-29
References in Nuxt setup

### LINKED ISSUE #94: References to features aren't detected in a Nuxt setup
Repro: ~~https://github.com/SVBadSeed/nuxt-steiger~~ https://github.com/feature-sliced/nuxt-steiger

```
┌ src/features/session/register
✘ This slice has no references. Consider removing it.
│
└ insignificant-slice (​https://github.com/feature-sliced/steiger/tree/master/packages/steiger-plugin-fsd/src/insignificant-slice​)

┌ src/features/session/logout
✘ This slice has no references. Consider removing it.
│
└ insignificant-slice (​https://github.com/feature-sliced/steiger/tree/master/packages/steiger-plugin-fsd/src/insignificant-slice​)

┌ src/features/session/login
✘ This slice has no references. Consider removing it.
│
└ insignificant-slice (​https://github.com/feature-sliced/steiger/tree/master/packages/steiger-plugin-fsd/src/insignificant-slice​)
```

There is a reference to these slices in `src/pages/home/ui/home.vue`

daniilsapa:
Resolves #94 

@illright 
The cause of the issue was not exactly in the Nuxt setup. The problem was that `compilerOption.paths` specified in config extensions (`"extends": "./.nuxt/tsconfig.json"`) were not resolved and thus path aliases like `@/`, `~/` didn't work. 

So such paths from extensions were applied to the FSD root.
```js
 {
     "~": [
        ".."
      ],
      "~/*": [
        "../*"
      ],
      "@": [
        ".."
      ],
      "@/*": [
        "../*"
      ],
}
```

I made changes to transform such paths from `../*` to `/*` and then apply them to the root.

Nuxt setup extends a default config from the `.nuxt` folder it helped us to uncover this issue.

Sorry for the delay!


packages/steiger-plugin-fsd/src/insignificant-slice/index.ts
```diff
@@ -93,3 +101,25 @@ async function traceSliceReferences(root: Folder) {
...
+function resolveRelativePathsInConfigExtensions(tsconfig: TsConfigExcerpts, configPath: string) {
+  if (!tsconfig || !tsconfig.extends || !tsconfig.compilerOptions) {
+    return tsconfig
+  }
+
+  const paths = tsconfig.compilerOptions?.paths ?? {}
+
+  const newPaths = Object.fromEntries(
+    Object.entries(paths).map(([key, values]) => {
+      return [key, values.map((value: string) => join(configPath, '..', value))]
```
illright, line 114: issue: the `'..'` part looks hard-coded for the Nuxt case. I don't fully understand how this function works and how it fixes the issue

illright (changes requested):
Could you also resolve the conflicts? Recently there's been a change to that file which did a similar thing, I wonder if your changes overlap with it or not

daniilsapa:
Got it, yes, let me resolve the conflicts first. Then I'll fix the function and also now I see some other flaws of my solution

packages/steiger-plugin-fsd/src/insignificant-slice/index.ts
illright, line 1: suggestion: the dependency resolution is also used in several other rules, it needs to be updated there as well, I think

daniilsapa: Yep, made the needed changes

daniilsapa:
@illright 

I fixed the first issue that we discussed in the morning by always transforming `compilerOptions.paths` to absolute, so now we're not afraid of losing context what Steiger is launched from the folder other than the one where the initial tsconfig is put.

Regarding the 2nd problem

https://stackblitz.com/edit/nuxt-starter-ap1c4y?file=README.md

**How to reproduce**

When `/src/features/session/logout/index` has **.js** extension. It reports this diagnostic, from which we can see that the file is noticed by the tool.
Screenshot 2024-11-11 at 18 29 00


When `/src/features/session/logout/index` has **.vue** extension. It reports the following diagnostic, from which we can see that the tool does not see that the file is imported in other files.
Screenshot 2024-11-11 at 18 29 44

Don't forget to wrap/unwrap the code inside `/src/features/session/logout/index` with `<script></script>` when changing extensions 


You can temporarily change fileExists, directoryExists parameters from these:

Screenshot 2024-11-11 at 18 38 27
to these
Screenshot 2024-11-11 at 18 37 45
(I made that change locally for insignificant-slice)

Then you'll be able to see how it goes through different possible options (like `.../logout.ts`, `.../logout.js`, `.../logout/index.ts`, `.../logout/index.js`, etc) and then fails when runs out of further options

daniilsapa:
resolveDependency from `.../steiger-plugin-fsd/src/_lib/resolve-dependency.ts` uses resolveImport from `@feature-sliced/filesystem`. It looks like we need to add some wrapper in resolveImport to add .vue extension case. Also, maybe we need to create space to gracefully add more custom extensions, but I'm not sure it's inside the score of the task, we can create a quick fix and an issue to design and implement a module responsible for custom extensions

illright (approved):

illright (approved):

Issue #30, opened 2024-06-09
Cover with E2E tests


illright:
Take some projects from FSD examples, pin a commit hash and run the linter on that commit. Then make fixes and run again to see the issues go away.

The test should run on all OSes on Node 18+

PR #124, opened 2024-11-12
Add the first integration test

illright:
First step towards #30


packages/steiger/src/app.ts
```diff
@@ -21,25 +21,22 @@ async function runRules({ vfs, rules }: { vfs: Folder; rules: Array<Rule> }) {
   const vfsWithoutGlobalIgnores = removeGlobalIgnoreFromVfs(vfs, getGlobalIgnores())
 
   const ruleResults = await Promise.all(rules.map((rule) => runRule(vfsWithoutGlobalIgnores, rule)))
-  return ruleResults.flatMap((r, ruleResultsIndex) => {
-    const { diagnostics } = r
-    if (diagnostics.length === 0) {
-      return []
-    }
```
illright, line 28: note: I removed this part while I was adding the sorting because I found several return statements to be confusing and because I believe this isn't that impactful from the performance perspective. The empty array will simply get processed as any other array, with no work done on it

daniilsapa (approved):
Great job! It's a good start with tests!

I only have one question and one optional suggestion.

Why have you stopped considering this test as an e2e test? Doesn't it test the application as a whole?

If you find it relevant, could you document the approach to adding more tests in the future? A short guide, like:
- Add posix test
- Run update-windows-snapshots
- Run tests
- Commit the results

but maybe in more details

illright:
> Why have you stopped considering this test as an e2e test? Doesn't it test the application as a whole?

I decided to call them "integration tests" because there are still aspects in a user journey that we are not testing — downloading the package from npm, colors in output, etc.

daniilsapa (approved):
Sorry for the delay! Thank you for the instructions! Now it looks good to go

PR #134, opened 2024-11-27
Added support of the steiger config for windows

mx-ryabov:
Hey! 
My version of tests looks a bit overwhelming, I also thought about a simple function that would take a platform parameter and depending on the value would prepend `C`:, but I'm not sure if that's the right way to do tests. However, if you think otherwise, I can rework it (or am open to any other suggestions)

### DIFF
--- a/.changeset/early-stingrays-appear.md
+++ b/.changeset/early-stingrays-appear.md
@@ -1,0 +1,5 @@
+---
+'steiger': minor
+---
+
+Added support of the steiger config for windows
--- a/.changeset/early-stingrays-appear.md
+++ b/.changeset/early-stingrays-appear.md
@@ -1,5 +1,5 @@
 ---
-'steiger': minor
+'steiger': patch
 ---
 
-Added support of the steiger config for windows
+Fix drive label handling in paths on Windows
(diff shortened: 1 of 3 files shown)


.changeset/early-stingrays-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'steiger': minor
+---
+
+Added support of the steiger config for windows
```
illright, line 5: suggestion: the support is already there, it's just buggy :)


```suggestion
Fix drive label handling in paths on Windows 
```

.changeset/early-stingrays-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'steiger': minor
```
illright, line 2: Also I think it should be a patch change, since it's a bug fix:
```suggestion
'steiger': patch 
```

illright (changes requested):
Thanks for making a PR! Sorry it took a while to get a review. The solution looks solid. The tests are a bit hard to parse, I think we can make them simpler by simply writing two `it`s and then skipping one or the other based on the platform. Our CI runs in both OSes, so we will cover everything that way

For reference: https://vitest.dev/api/#test-runif

mx-ryabov:
I considered a similar to yours approach but I thought about the case when somebody on a unix machine could break something for windows and this developer won't find out it until the CI checks passed.
But anyway all your comments are resolved, let me know if you need something else. Thank you!

illright (approved):
Thanks for addressing the comments! Nothing else on my side, great job, let's merge.

As for your concern about devs not finding out until CI — there's a lot of places where that can happen, and that's okay, that's why we have CI on several OSes, to alert devs that something might not work on someone else's machine

PR #139, opened 2024-12-05
fix: make the version number correct

### LINKED ISSUE #127: Steiger gives you "unknown" when you ask for its version
Screenshot 2024-11-16 at 14 56 01

illright:
Closes #127. I wanted to go fancy with colors in the output but realized that it doesn't actually improve much. So went with as much of a default presentation as possible :)

image

### DIFF
--- a/packages/steiger/src/cli.ts
+++ b/packages/steiger/src/cli.ts
@@ -11,10 +11,27 @@ import { fromError } from 'zod-validation-error'
 import { cosmiconfig } from 'cosmiconfig'
 
 import { linter } from './app'
-import { processConfiguration } from './models/config'
+import { processConfiguration, $plugins } from './models/config'
 import { applyAutofixes } from './features/autofix'
 import fsd from '@feature-sliced/steiger-plugin'
 import type { Diagnostic } from '@steiger/types'
+import packageJson from '../package.json'
+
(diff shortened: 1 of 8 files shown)


packages/steiger/src/cli.ts
```diff
@@ -55,28 +72,22 @@ const yargsProgram = yargs(hideBin(process.argv))
   })
   .help('help', 'display help message')
   .alias('help', 'h')
-  .version()
+  .version(
+    [
+      packageJson.version,
+      $plugins
```
daniilsapa, line 78: We used to define and export getter functions from the config module before (like `getEnabledRules`, and `getGlobalIgnores`). Does it make sense in the case of plugins too for consistency?

daniilsapa:
@illright 
Looks great. I don't have any suggestions except for a single tiny one that I left a comment for. Thank you

packages/steiger/src/cli.ts
daniilsapa: ...so it just helps us to keep the codebase less dependent on the store solution and keep the code that interacts with effector mainly inside `models/config`

illright: I don't like getters for the sake of getters. In this case, it doesn't matter architecturally if we do `$plugins.getState()` or `getPlugins()`, because we aren't using Effector-specific functionality either way.

As for consistency, that is not the only store that's exported there, so we didn't have consistency there to begin with :)

PR #141, opened 2024-12-07
Handle an edge case with well-known ignores when applying exclusions

### LINKED ISSUE #138: Steiger reports insignificant-slice diagnostics on disabled part of file system
Additionally, it can report issues on a layer.

Part of my config:
Screenshot 2024-12-04 at 19 13 03

Diagnostics:
Screenshot 2024-12-04 at 19 12 53

daniilsapa:
Closes #138

This was a subtle one. The problem was that the folders that were disabled in config but still had diagnostics reported on them in my local env contained `.DS_Store files`. `.DS_Store` files prevented a folder from being excluded from VFS. The reason is that minimatch apparently has some registry of ignored files and it always returns false for them. It becomes a problem when we apply inverted glob groups (`not(...)`) as we invert the result when matching such a group and end up including those files/folders

Let me know if I failed to clearly articulate the problem and you need more explanations

### DIFF
--- a/packages/steiger/src/shared/globs/glob-matchers.ts
+++ b/packages/steiger/src/shared/globs/glob-matchers.ts
@@ -40,3 +40,14 @@ export function createFilterAccordingToGlobs({ inclusions, exclusions }: ApplyGl
 
   return filterAccordingToGlobs
 }
+
+/**
+ * Checks if a file/folder is a well-known ignore.
+ *
+ * If a path does not match the "any file/folder" pattern, it is considered a well-known ignore.
+ * It looks like minimatch has some registry of such files, and it returns false for them regardless of the pattern.
+ * It's the case for .DS_Store on macOS. Maybe there are other cases.
+ * */
+export function isWellKnownIgnore(path: string) {
+  return !minimatch(path, '**')
+}
--- a/.changeset/eleven-numbers-leave.md
+++ b/.changeset/eleven-numbers-leave.md
@@ -1,0 +1,5 @@
+---
+'steiger': patch
+---
+
+Handle an edge case with well-known ignores (like .DS_Store on macOS) when applying glob exclusions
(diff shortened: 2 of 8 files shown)


packages/steiger/src/shared/globs/glob-matchers.ts
```diff
@@ -40,3 +40,14 @@ export function createFilterAccordingToGlobs({ inclusions, exclusions }: ApplyGl
 
   return filterAccordingToGlobs
 }
+
+/**
+ * Checks if a file/folder is a well-known ignore.
+ *
+ * If a path does not match the "any file/folder" pattern, it is considered a well-known ignore.
+ * It looks like minimatch has some registry of such files, and it returns false for them regardless of the pattern.
```
illright, line 48: note: I think what's happening here is that hidden files (those whose name starts with a dot) are not matched by globs by default. That's how terminal shells handle globs too. 

issue: I don't think ignoring hidden files it's the right approach here, since there are legitimate reasons for why you might want to make a source code file hidden.

suggestion: I'd rather investigate why these files are picked up at all. I was under the impression that we automatically disregard the files that are in .gitignore. If we don't, then we should, I think.

illright (changes requested):
Let's try to find another solution, because this solution might actually produce unexpected behavior for user-created hidden files (i.e. `.eslintrc.js`, contrived example, but still). Maybe we can even discuss it in the issue rather than in the PR

packages/steiger/src/shared/globs/glob-matchers.ts
daniilsapa: Oh, yeah, it was so obvious to me that it should have ignored only default hidden FS-specific files that I didn't even  checked how it behaves for dot files as such. Yes, it turns out this happens with any dot files, not just a few specific ones. Given these new circumstances, yes, that's not the right approach at all.

Yes, I'll investigate why the files that are in .gitignore are not disregarded by default

daniilsapa:
I rolled my previous changes back. It turned out you need to supply globby's `isGitIgnored` with a path to the root of a git repo. It doesn't search it recursively. Adding that function solved the problem

illright (approved):
Thanks for finding the issue and fixing it :)

PR #145, opened 2024-12-09
Add the workflow to publish package previews

illright:
This is an attempt to add pkg.pr.new

### DIFF
--- a/.github/workflows/pkg-pr-new.yaml
+++ b/.github/workflows/pkg-pr-new.yaml
@@ -1,0 +1,48 @@
+name: Publish Approved Pull Requests to pkg.pr.new
+on:
+  pull_request_review:
+    types: [submitted]
+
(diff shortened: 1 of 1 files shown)


daniilsapa (approved):
It looks good so far, but I think we need a real run to see if it runs as expected

illright:
@daniilsapa could you approve it again please? it only runs when someone approves, and the last run failed (I think I fixed it now)

daniilsapa:
I see some warnings called out by CodeQL. They suggested us adding permissions for the workflow to improve security and ensure it does not touch what it mustn't.

e.g.
```
name: Publish Approved Pull Requests to pkg.pr.new
permissions:
  contents: read
  pull-requests: write
```

daniilsapa (approved):
🤞

PR #146, opened 2024-12-10
Upgrade dependencies

illright:
This was originally intended as a throw-away PR to try and break formatting and linting checks to verify that they work, but I noticed some config issues and decided to fix them, as well as upgrade all dependencies, something I wanted to do for a while

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,4 @@
 pnpm-lock.yaml
 .changeset/*
+packages/**
+tooling/**
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,2 @@
 pnpm-lock.yaml
 .changeset/*
-packages/**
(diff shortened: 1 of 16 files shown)


daniilsapa (commented):
@illright 
Looks great

I have only 2 questions:
- Why did you add /packages and /tooling to .prettierignore?
- I don't fully understand what caused the change of NodeJs version 

(it's just interest, not complaints)

illright:
I added them to the ignore section of the root Prettier command because packages are responsible for their own formatting. That way, when we run `turbo format`, the commands don't overlap in the files that they scan.

As for the Node version change — we test on Node 18, so I updated the type definitions to include the APIs that are present in Node 18 so that we know upfront with TypeScript when something is not available in Node 18, but is available in a later version.

PR #149, opened 2024-12-16
Make locations visible (Intellij)

### LINKED ISSUE #140: Diagnostic locations look unreadable in IntelliJ in dark mode
As seen in #138:

It's also a fine time to replace `chalk` with `picocolors` and maybe `dim()` will look better here

daniilsapa:
Resolves #140

@illright 
I replaced `chalk` with `picocolors` as you suggested in the issue description.

Here you can compare the diagnostics (I used `dim` method)

Light theme before/after:
Screenshot 2024-12-16 at 15 46 56
Screenshot 2024-12-16 at 15 44 37

Dark theme before/after:
Screenshot 2024-12-16 at 15 46 42
Screenshot 2024-12-16 at 15 45 03

### DIFF
--- a/packages/pretty-reporter/src/format-single-diagnostic.ts
+++ b/packages/pretty-reporter/src/format-single-diagnostic.ts
@@ -1,19 +1,18 @@
 import { relative } from 'node:path'
 import figures from 'figures'
 import terminalLink from 'terminal-link'
-import picocolors from 'picocolors'
+import pc from 'picocolors'
 import type { Diagnostic } from '@steiger/types'
 
 export function formatSingleDiagnostic(d: Diagnostic, cwd: string): string {
-  const x = d.severity === 'error' ? picocolors.red(figures.cross) : picocolors.yellow(figures.warning)
-  const s = picocolors.reset(figures.lineDownRight)
-  const bar = picocolors.reset(figures.lineVertical)
-  const e = picocolors.reset(figures.lineUpRight)
-  const message = picocolors.reset(d.message)
-  const autofixable =
-    d.fixes !== undefined && d.fixes.length > 0 ? picocolors.green(`${figures.tick} Auto-fixable`) : null
-  const location = picocolors.dim(formatLocation(d.location, cwd))
-  const ruleName = picocolors.blue(terminalLink(d.ruleName, d.getRuleDescriptionUrl(d.ruleName).toString()))
+  const x = d.severity === 'error' ? pc.red(figures.cross) : pc.yellow(figures.warning)
+  const s = pc.reset(figures.lineDownRight)
+  const bar = pc.reset(figures.lineVertical)
+  const e = pc.reset(figures.lineUpRight)
+  const message = pc.reset(d.message)
+  const autofixable = d.fixes !== undefined && d.fixes.length > 0 ? pc.green(`${figures.tick} Auto-fixable`) : null
+  const location = pc.dim(formatLocation(d.location, cwd))
+  const ruleName = pc.blue(terminalLink(d.ruleName, d.getRuleDescriptionUrl(d.ruleName).toString()))
 
   return `
 ${s} ${location}
--- a/packages/pretty-reporter/src/format-single-diagnostic.ts
+++ b/packages/pretty-reporter/src/format-single-diagnostic.ts
@@ -11,7 +11,7 @@ export function formatSingleDiagnostic(d: Diagnostic, cwd: string): string {
   const e = pc.reset(figures.lineUpRight)
   const message = pc.reset(d.message)
   const autofixable = d.fixes !== undefined && d.fixes.length > 0 ? pc.green(`${figures.tick} Auto-fixable`) : null
-  const location = pc.dim(formatLocation(d.location, cwd))
+  const location = pc.underline(formatLocation(d.location, cwd))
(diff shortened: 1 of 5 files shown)


daniilsapa:
The font looks a bit lighter, so it's a bit more readable in the dark theme now. Do you think it's enough?

daniilsapa:
Also, here's the difference for VS Code

Dark theme before/after:
Screenshot 2024-12-16 at 15 55 40
Screenshot 2024-12-16 at 15 58 09


Light theme before/after:
Screenshot 2024-12-16 at 15 56 18
Screenshot 2024-12-16 at 15 57 26

illright:
I like it more in the dark mode of IntelliJ, but I like it less in the light mode of VS Code, seems a bit too pale there. I'll also play around some more. My main target here is sufficient contrast between the background and the text, we can use https://webaim.org/resources/contrastchecker/ to check

daniilsapa:
Yeah, I'll play around with the contrast a bit and send you some screenshots. I also thought about coloring the location according to severity, e.g. dark yellow/red for warn/error.

daniilsapa:
Here you can see the variant of locations colored according to the severity:

Error:
Screenshot 2024-12-17 at 00 46 47
Screenshot 2024-12-17 at 00 48 02

Warning:
Screenshot 2024-12-17 at 00 48 35
Screenshot 2024-12-17 at 00 48 53

illright:
Nah, this ain't it either. The red color is too attention-grabbing, we don't want the path to draw attention away from the error text

daniilsapa:
Also, `#757575` looks quite contrasty against both white and black backgrounds. But unfortunately, picocolors does not allow us to set arbitrary hex values for colors. Additionally, we could make the font bolder, it helps with readability a bit

**Intellij**
Screenshot 2024-12-17 at 01 31 37
Screenshot 2024-12-17 at 01 31 52


**VS Code**
Screenshot 2024-12-17 at 01 32 59
Screenshot 2024-12-17 at 01 33 29

illright:
Just played around with it myself, seems like `#757575` is `pc.dim`, which doesn't pass the contrast check against the default dark theme of VS Code. `pc.gray`, however, is `#919191`, and that's better

https://webaim.org/resources/contrastchecker/?fcolor=919191&bcolor=1E1E1E

packages/pretty-reporter/src/format-single-diagnostic.ts
```diff
@@ -1,18 +1,19 @@
 import { relative } from 'node:path'
 import figures from 'figures'
 import terminalLink from 'terminal-link'
-import chalk from 'chalk'
+import picocolors from 'picocolors'
```
illright, line 4: suggestion (non-blocking): I think the convention is to import is as `pc`, also helps make the code shorter
```suggestion
import pc from 'picocolors'
```

daniilsapa:
> Just played around with it myself, seems like `#757575` is `pc.dim`, which doesn't pass the contrast check against the default dark theme of VS Code. `pc.gray`, however, is `#919191`, and that's better

Yes, `pc.gray` looks better for the default dark theme of VS Code. But it does not look readable in dark themes of IntelliJ. Also, these colors have different values in different themes. Here's a table where you can see all color values and their contrast rates. `pc.gray` has better overall results but looks very unreadable on dark themes in IntelliJ. `pc.dim` is bad everywhere, but I would say it's "average bad", it has values that are a bit lower than the pass criteria almost everywhere.

| Theme                       | Dim (actual color, contrast rate) | Gray (actual color, contrast rate) |
|-----------------------------|----------------------------------|-----------------------------------|
| VS Code Dark (bg #1e1e1e)   | #737373, 3.51:1                  | #919191, 5.28:1 - pass            |
| VS Code Dark Modern (bg #181818) | #707070, 3.58:1               | #828282, 4.62:1 - pass            |
| VS Code Light (bg #ffffff)  | #a0a0a0, 2.61:1                  | #666666, 5.74:1 - pass            |
| VS Code Light Modern (bg #f8f8f8) | #a0a0a0, 2.46:1              | #6C6C6C, 4.94:1 - pass            |
| Intellij Dark (bg #1e1f22)  | #757679, 3.62:1                  | #4a4c51, 1.91:1                   |
| IntelliJ Dracula (bg #2b2b2b) | #858688, 3.88:1                | #4f5156, 1.78:1                   |
| Intellij Light (bg #ffffff) | #7f7f7f, 4:1                    | #6d707d, 4.92:1 - pass            |

I'll try to do something with `pc.gray` in IntelliJ

packages/pretty-reporter/src/format-single-diagnostic.ts
daniilsapa: Yep, I'll shorten the variable name

daniilsapa:
I've reviewed eslint logic for printing diagnostics and it's nothing crazy. They just use `chalk.dim` for the line:column part and the rule id.



Screenshot 2024-12-24 at 17 05 39

So, I eventually decided to go with underlined text so as not to overengineer the solution. Here you can see what it looks like

Screenshot 2024-12-24 at 17 11 00
Screenshot 2024-12-24 at 17 11 25

illright:
I see there are also some empty files left - over, are these important or can we delete them?

daniilsapa:
Oh, yeah, I've already deleted them, those 2 accidentally got committed

illright (approved):
Great, let's merge this then!

PR #150, opened 2024-12-17
Add a workflow to print bundle size

### LINKED ISSUE #148: Add a bundle size check to PRs
To see how each PR affects the package sizes

illright:
Closes #148. I wanted to find an existing solution, but they all suck in some ways, so I just decided to write my own one

### DIFF
--- a/.github/workflows/bundle-size-trusted.yaml
+++ b/.github/workflows/bundle-size-trusted.yaml
@@ -63,7 +63,7 @@ jobs:
             return report;
 
       - name: Post the report to the PR
-        uses: thollander/actions-comment-pull-request@v3
+        uses: thollander/actions-comment-pull-request@24bffb9b452ba05a4f3f77933840a6a841d1b32b # v3.0.1
         with:
           message: ${{ steps.create-report.outputs.result }}
           pr-number: ${{ github.event.workflow_run.pull_requests.number }}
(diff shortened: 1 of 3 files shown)


illright:
The trusted part of the workflow won't work until we merge, so we'll have to wait with testing

daniilsapa:
Wonderful job! I'm wondering why you had to split it into two parts.

illright:
It has to do with security. The untrusted part of the workflow runs on the PR branch, and theoretically, someone could submit a malicious PR that changes the workflow to do something bad. We wouldn't want to run this workflow with a token that has write access, so GitHub runs it with the read access by default. However, that means we don't have permissions to "write" comments on PRs.

In this setup, we run a read-only workflow in untrusted land and then publish a plain-text artifact that can then be safely downloaded and parsed in the trusted land without executing any unsafe code

daniilsapa:
Got it. I'm not a big pro in GH Actions, so it may sound stupid to you, but can't we just do all the needed jobs in the "trusted land"? So we don't split the workflow into 2 parts and keep it simpler

illright:
We can't, because the reason why the trusted land is trusted is because there is no external code from the PR :) We want to run some safe operations on external code in untrusted land (i.e. counting the bundle size) and then publish a comment in trusted land

daniilsapa:
Got it, then yes, let's end the "Guardians of the Trusted Land" saga 😂 and merge it

guardians-of-the-trusted-land

daniilsapa (approved):

PR #151, opened 2024-12-19
Move getRuleDescriptionUrl to plugins

### LINKED ISSUE #131: Move getRuleDescriptionUrl from steiger to steiger-plugin
`getRuleDescriptionUrl` function must be part of a plugin since plugins will have their own repositories and therefore different locations of rule descriptions.

daniilsapa:
Closes #131

I was not sure about one thing: should `getRuleDescriptionUr`l be an optional plugin method or not? I decided to make it required because otherwise, a diagnostic is not that useful. What do you think?

### DIFF
--- a/packages/steiger/src/app.ts
+++ b/packages/steiger/src/app.ts
@@ -20,7 +20,7 @@ async function runRules({ vfs, rules }: { vfs: Folder; rules: Array<Rule> }) {
     }
 
     const ruleName = rules[ruleResultsIndex].name
-    const pluginProvidedTheRule = getPluginByRuleName(ruleName)
+    const ruleSourcePlugin = getPluginByRuleName(ruleName)
     const severities = calculateFinalSeverities(
       vfsWithoutGlobalIgnores,
       ruleName,
--- a/packages/steiger/src/app.ts
+++ b/packages/steiger/src/app.ts
@@ -4,18 +4,10 @@ import type { Config, Folder, Rule } from '@steiger/types'
 
 import { scan, createWatcher } from './features/transfer-fs-to-vfs'
 import { defer } from './shared/defer'
-import { $enabledRules, getEnabledRules, getGlobalIgnores } from './models/config'
+import { $enabledRules, getEnabledRules, getGlobalIgnores, getPluginByRuleName } from './models/config'
 import { runRule } from './features/run-rule'
 import { removeGlobalIgnoreFromVfs } from './features/remove-global-ignores-from-vfs'
 import { calculateFinalSeverities } from './features/calculate-diagnostic-severities'
-
-// TODO: make this part of a plugin
-function getRuleDescriptionUrl(ruleName: string) {
-  const withoutNamespace = ruleName.split('/')[1]
-  return new URL(
-    `https://github.com/feature-sliced/steiger/tree/master/packages/steiger-plugin-fsd/src/${withoutNamespace}`,
-  )
-}
 
 async function runRules({ vfs, rules }: { vfs: Folder; rules: Array<Rule> }) {
   const vfsWithoutGlobalIgnores = removeGlobalIgnoreFromVfs(vfs, getGlobalIgnores())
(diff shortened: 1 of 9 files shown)


packages/steiger/src/app.ts
```diff
@@ -28,6 +20,7 @@ async function runRules({ vfs, rules }: { vfs: Folder; rules: Array<Rule> }) {
     }
 
     const ruleName = rules[ruleResultsIndex].name
+    const pluginProvidedTheRule = getPluginByRuleName(ruleName)
```
illright, line 23: issue (non-blocking): this reads like a sentence to me, which is a bit confusing for a variable name

suggestion: how about this:
```suggestion
    const ruleSourcePlugin = getPluginByRuleName(ruleName)
```

packages/steiger/src/models/config/index.ts
```diff
@@ -48,6 +48,11 @@ export function getEnabledRules() {
   return $enabledRules.getState()
 }
 
+export function getPluginByRuleName(ruleName: string) {
+  // Always returns a plugin because if a rule exists than the plugin provided it must exist too
+  return $plugins.getState().find((plugin) => plugin.ruleDefinitions.some((rule) => rule.name === ruleName))!
```
illright, line 53: issue: not a big fan of the `!` operator, especially in a case like this. If it does somehow fail to return the plugin, we might have an error that's harder to debug

suggestion: let's throw a descriptive error if the search failed. That way, if it ever happens, we'll know exactly what didn't match our expectations
```suggestion
  const sourcePlugin = $plugins.getState().find((plugin) => plugin.ruleDefinitions.some((rule) => rule.name === ruleName))
  if (sourcePlugin === undefined) {
    throw new Error(`Expected to find a source plugin for the rule "${ruleName}", but didn't`)
  }
  
  return sourcePlugin
```

packages/steiger/src/models/config/validate-config.spec.ts
```diff
@@ -8,11 +8,12 @@ import { isPlugin } from './raw-config'
 // but with the check method expected to be just any function.
 // The reason why it's needed is that the original check method's .toString method returns [Function: check],
 // but after the validation it changes to [Function: anonymous] and fails the test.
-function expectCheckToBeAnyFunction(config: Config<Array<Rule>>) {
+function expectFunctionsToBeAnyFunction(config: Config<Array<Rule>>) {
```
illright, line 11: issue: now this name is confusing enough to me that even the comment before it doesn't quite cut it :D

suggestion: how about, instead of checking for equality, we simply make it check that the call to `.parse` doesn't throw? That's what we're trying to test, after all, and I can't come up with a way to neatly do equality checking either

packages/steiger/src/models/config/index.ts
```diff
@@ -48,6 +48,11 @@ export function getEnabledRules() {
   return $enabledRules.getState()
 }
 
+export function getPluginByRuleName(ruleName: string) {
```
illright, line 51: praise: now that's a nice getter in my book :)

illright (changes requested):

illright:
About optionality — I think I would prefer it to be optional to lower the entry barrier for writing rules. It feels strange to link to documentation that doesn't yet exist, so we want to let people add the docs when it's convenient. If they don't have docs, they can always make the error message more descriptive

packages/steiger/src/app.ts
daniilsapa: Yep, I don't mind, makes sense

packages/steiger/src/models/config/index.ts
daniilsapa: I agree, this is better

packages/steiger/src/models/config/validate-config.spec.ts
daniilsapa: Lol, I don't even know why I haven't done that in the first place and ended up inventing a square wheel 🤷‍♂️

daniilsapa:
Actually, optionality makes sense in this case. I think I'll change it a bit so that function can provide either a link or a short description for a diagnostic. Is that something you thought too?

illright:
> I'll change it a bit so that function can provide either a link or a short description for a diagnostic. Is that something you thought too?

No, I meant that the explanation can be given in the error message itself. How would you like to display this short description that a function might return?

daniilsapa:
I was thinking about a short description that could be returned by the `getRuleDescription` function. Then we could display that instead of a link. But I looked at our diagnostics again and saw that a short description instead of a link is a bit of nonsense. So never mind...

daniilsapa:
I made `getRuleDescriptionUrl` optional

illright (approved):
All good!

PR #154, opened 2024-12-22
Add an exception to the `insignificant-slice` rule when the only usage site is the App layer

### LINKED ISSUE #153: Implement an exception for the App layer in `insignificant-slice`
Someone in the chat had an issue — they have `widgets/sidebar` that's only used in `app/layouts`. Steiger complains that the `sidebar` slice is insignificant and should be moved to `app/layouts`, but the App layer shouldn't have UI.

illright:
Closes #153


daniilsapa (approved):
Thank you, approved, it's a good thing to include in 0.5.4

PR #158, opened 2024-12-30
Refactor tests

### LINKED ISSUE #128: Make test cases closer to real use of the tool
Many test cases in Steger assume that the user's project is at the root of the filesystem, which is virtually never the case, but can lead to subtle bugs leaking into the released code. Thus, we need to make sure that the FS mocks used in the test cases are nested at some level.

daniilsapa:
Closes #128

### DIFF
--- a/packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
+++ b/packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
@@ -12,7 +12,7 @@ vi.mock('tsconfck', async (importOriginal) => {
           compilerOptions: {
             baseUrl: '/src/',
             paths: {
-              '@/*': ['./*'],
+              '@/*': ['/users/user/project/src/*'],
             },
           },
         },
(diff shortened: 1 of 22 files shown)


packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
```diff
@@ -12,7 +12,7 @@ vi.mock('tsconfck', async (importOriginal) => {
           compilerOptions: {
             baseUrl: '/src/',
             paths: {
-              '@/*': ['./*'],
+              '@/*': ['/users/user/project/src/*'],
```
illright, line 15: issue (non-blocking): this is slightly incorrect from the perspective of tsconfig semantics. When `baseUrl` is set, `paths` are computed relative to it. Therefore, it's a little strange when `baseUrl` and the alias resolve to completely different locations

suggestion: 
```suggestion
            baseUrl: '/users/user/project/src/',
            paths: {
              '@/*': ['./*'],
```

packages/steiger-plugin-fsd/src/import-locality/index.spec.ts
```diff
@@ -6,7 +6,9 @@ import importLocality from './index.js'
 vi.mock('tsconfck', async (importOriginal) => {
   return {
     ...(await importOriginal<typeof import('tsconfck')>()),
-    parse: vi.fn(() => Promise.resolve({ tsconfig: { compilerOptions: { paths: { '@/*': ['/*'] } } } })),
+    parse: vi.fn(() =>
+      Promise.resolve({ tsconfig: { compilerOptions: { paths: { '@/*': ['/users/user/project/src/*'] } } } }),
```
illright, line 10: note: but here, for instance, it's fine because the alias was already an absolute path, so it's okay to just replace it

illright (approved):
Great, thanks! I have one optional suggestion, but feel free to merge if you don't want to bother with it

packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
daniilsapa: Yes, you're completely right. I'm surprised that it hasn't broken anything

PR #159, opened 2024-12-30
Prevent links to rule descriptions being broken in some terminals

### LINKED ISSUE #73: `terminal-link` doesn't work anymore?
image

The links should not be written out like that, they're ugly and long

illright:
Closes #73. It's not the best it can be, but until we work out a short domain, that should suffice.
image


daniilsapa (approved):
It looks good to go. Maybe not perfect as you admitted, but good enough, thank you!

PR #160, opened 2025-01-02
Collapse output to 20 items

### LINKED ISSUE #51: Collapse the output to 20 diagnostics maximum
Ideally we should do it cleverly — if there are more diagnostics than the limit, we should make sure that we show as many unique issues as possible. Usually that means taking a diagnostic from each rule until we hit the limit and then printing it out, but still printing diagnostics from the same rule together.

Also we should mention in the footer of the output that the actual amount of diagnostics is larger, it's just that X of them were hidden

daniilsapa:
Closes #51 

The solution is still raw and it's an early PR, so you can see the general approach and we can clarify some details.

1. When I was working on this logic I came across a thought that we probably need to make the number we collapse diagnostics to configurable and introduce a new config object type for such things that configure Steiger itself (it is very likely that other options will appear pretty soon) and not severities/options for rules. It seems it's better to introduce such a new object type earlier, e.g. in the next minor update.
2. Should we introduce severity to consideration when we collapse diagnostics? Looks like we need to count error first because it's the thing that breaks CIs or local checks. It would be quite annoying if you do not see an error because the quota is exceeded with warnings
3. It looks like `runRules` from `app.ts` needs to return some `Report` interface, so it contains not just diagnostics but also some meta info about them. We need this to display the number of shown vs hidden diagnostics etc. other fields may appear in the future.

### DIFF
--- a/packages/pretty-reporter/src/collapse-diagnostics.test.ts
+++ b/packages/pretty-reporter/src/collapse-diagnostics.test.ts
@@ -1,141 +1,90 @@
 import { describe, expect, it } from 'vitest'
-import { trimDiagnosticsToMeetQuota } from './collapse-diagnostics.js'
+import { Diagnostic } from '@steiger/types'
 
-const defaultLocation = {
-  path: '/users/user/file',
-}
+import { collapseDiagnostics } from './collapse-diagnostics.js'
 
-const defaultSeverity = 'error' as 'warn' | 'error'
+function makeDummyDiagnostic(rule: string, messageNum: number, severity?: 'warn' | 'error'): Diagnostic {
+  const defaultLocation = {
+    path: '/users/user/file',
+  }
+  const defaultSeverity = 'error' as 'warn' | 'error'
 
-// dummy rules
-
-const rule1Message1 = {
-  message: 'First rule, first message',
-  ruleName: 'rule-1',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule1Message2 = {
-  message: 'First rule, second message',
-  ruleName: 'rule-1',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule1Message3 = {
-  message: 'First rule, third message',
-  ruleName: 'rule-1',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule2Message1 = {
-  message: 'Second rule, first message',
-  ruleName: 'rule-2',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule2Message2 = {
-  message: 'Second rule, second message',
-  ruleName: 'rule-2',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule3Message1 = {
-  message: 'Third rule, first message',
-  ruleName: 'rule-3',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule3Message2 = {
-  message: 'Third rule, second message',
-  ruleName: 'rule-3',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule4Message1 = {
-  message: 'Forth rule, first message',
-  ruleName: 'rule-4',
-  location: defaultLocation,
-  severity: defaultSeverity,
-}
-
-const rule4Message2 = {
-  message: 'Forth rule, second message',
-  ruleName: 'rule-4',
-  location: defaultLocation,
-  severity: defaultSeverity,
+  return {
+    message: `${rule}, message ${messageNum}`,
+    ruleName: rule,
+    location: defaultLocation,
+    severity: severity ?? defaultSeverity,
+  }
 }
 
 describe('trimDiagnosticsToMeetQuota', () => {
   it('should return the same diagnostics if they are below the quota', () => {
     const diagnosticPerRule = [
-      [rule1Message1, rule1Message2, rule1Message3],
-      [rule2Message1, rule2Message2],
+      new Array(3).fill(0).map((_, i) => makeDummyDiagnostic('rule-1', i + 1)),
+      new Array(2).fill(0).map((_, i) => makeDummyDiagnostic('rule-2', i + 1)),
     ]
 
-    const result = trimDiagnosticsToMeetQuota(diagnosticPerRule, 10)
+    const result = collapseDiagnostics(diagnosticPerRule)
 
     expect(result).toEqual(diagnosticPerRule)
   })
 
-  it('should return no diagnostics if the quota is 0', () => {
-    const diagnosticPerRule = [
-      [rule1Message1, rule1Message2, rule1Message3],
-      [rule2Message1, rule2Message2],
-    ]
-
-    const result = trimDiagnosticsToMeetQuota(diagnosticPerRule, 0)
-
-    expect(result).toEqual([[], []])
-  })
-
   it('should spread the quota evenly between rules', () => {
     const diagnosticPerRule = [
-      [rule1Message1, rule1Message2, rule1Message3],
-      [rule2Message1, rule2Message2],
+      new Array(20).fill(0).map((_, i) => makeDummyDiagnostic('rule-1', i + 1)),
+      new Array(20).fill(0).map((_, i) => makeDummyDiagnostic('rule-2', i + 1)),
     ]
 
-    const result = trimDiagnosticsToMeetQuota(diagnosticPerRule, 2)
+    const result = collapseDiagnostics(diagnosticPerRule)
 
-    expect(result).toEqual([[rule1Message1], [rule2Message1]])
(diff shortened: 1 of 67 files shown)


illright:
Interesting considerations, thanks for the early update! I looked at the code and realized that I forgot to mention one important detail in the issue description.

I originally envisioned this as a purely cosmetic thing rather than a linter feature, and as such, I planned that this functionality would be added to the pretty reporter rather than the linter core. This way, all reporters get the same full diagnostics, but some reporters might choose to hide some.

The inspiration for this feature came from [Biome](https://biomejs.dev/), they also have a 20-diagnostic limit. Here's how they show it:
image

That's why I don't think we should make this thing configurable or even disable-able. If one wants to see all diagnostics, I think it's not because they want to read all of them individually. For mass-reading of diagnostics, the JSON reporter is probably better. I suggest we write "X diagnostics were hidden, run with `--reporter json` to see all diagnostics" or something.

The severity consideration is a very good point, I agree that we probably want to sort the errors first and then, if there's space left, print warnings too. If there's a rule that has produced only warnings, then we only print its diagnostics if all the errors have been printed.

daniilsapa:
Oh, yeah, actually it makes sense to make it a cosmetic thing and move `pretty-reporter`. It would also make the implementation easier and the resulting codebase less complex. Do you like the general approach? If so, I will move the logic to pretty-reporter, add severity to consideration, and refactor it a bit.

illright:
Overall the code looks good, I only have minor concerns about over-abstracting — such as having an entire feature for this, and also making the quota a modifiable parameter. Seems like the code could be simpler by keeping it all in one function

daniilsapa:
@illright 
I've moved the solution to pretty-reporter. Also, you can see the output here:

Screenshot 2025-01-06 at 15 51 11

Let me know if you think we must rephrase/change it somehow

daniilsapa:
The bundle size pipeline fails when the git branch contains slashes. I'll fix the problem as part of this issue

Screenshot 2025-01-06 at 16 51 03

packages/pretty-reporter/src/collapse-diagnostics.test.ts
```diff
@@ -0,0 +1,141 @@
+import { describe, expect, it } from 'vitest'
+import { trimDiagnosticsToMeetQuota } from './collapse-diagnostics.js'
+
+const defaultLocation = {
+  path: '/users/user/file',
+}
+
+const defaultSeverity = 'error' as 'warn' | 'error'
+
+// dummy rules
```
illright, line 10: suggestion (non-blocking): how about we instead write a function to generate dummy diagnostics with a simple index in the message and then call it in the test setup? This way we can even test with quota = 20

```ts
function makeDiagnostic(ruleName: string, index: number): PartialDiagnostic
```
```ts
const diagnosticPerRule = [
  new Array(15).fill(0).map((_, index) => makeDiagnostic('rule-1', index)),
  new Array(10).fill(0).map((_, index) => makeDiagnostic('rule-2', index)),
]
```

illright: And in the tests that actually drop some diagnostics, we can assert by doing `diagnosticPerRule[0].slice(0, 5)`. This way we also test that it doesn't change the order of diagnostics within one rule

packages/pretty-reporter/src/collapse-diagnostics.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    const diagnosticPerRule = [
+      [rule1Message1, rule1Message2, rule1Message3],
+      [rule2Message1, rule2Message2],
+      [rule3Message1, rule3Message2],
+      [rule4Message1, rule4Message2],
+    ]
+
+    const result = trimDiagnosticsToMeetQuota(diagnosticPerRule, 3)
+
+    expect(result).toEqual([[rule1Message1], [rule2Message1], [rule3Message1]])
```
illright, line 119: note: for some reason, this test fails for me locally because `result` also includes an empty array at the end

illright (changes requested):
Thanks for noticing the bundle size workflow issue, my mistake! I think we can replace everything that isn't `/[0-9a-zA-Z-._]/` ([ref](https://superuser.com/a/748264)) with `-` to overcome this.

The solution is good overall, but slightly hard to understand, I made some suggestions to improve clarity, let me know what you think

packages/pretty-reporter/src/index.ts
```diff
@@ -22,8 +28,8 @@ export function formatPretty(diagnostics: Array<Diagnostic>, cwd: string) {
       .filter(Boolean)
       .join(' and ')
 
-  const autofixable = diagnostics.filter((d) => (d.fixes?.length ?? 0) > 0)
-  if (autofixable.length === diagnostics.length) {
+  const autofixable = collapsedDiagnostics.filter((d) => (d.fixes?.length ?? 0) > 0)
+  if (autofixable.length === collapsedDiagnostics.length) {
```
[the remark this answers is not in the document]
daniilsapa: 100%, thank you for catching it, I actually thought the same but then forgot to revert this part of my changes

packages/pretty-reporter/src/collapse-diagnostics.ts
```diff
@@ -0,0 +1,51 @@
...
+    // If the bucket already has the quota distributed for it or does not contain any items, skip it
+    if (assignedQuotaForBucket < numOfItemsInBucket && numOfItemsInBucket !== 0) {
+      quotaPerBucket[i] += 1
+      remainingQuota -= 1
+    }
+
+    // If it ran out of the items earlier than the quota, break
+    if (allItems === quota - remainingQuota) {
+      break
+    }
```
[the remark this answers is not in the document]
daniilsapa: Yes, your suggestion looks much better than that condition

packages/pretty-reporter/src/collapse-diagnostics.test.ts
daniilsapa: Yes, you're right, it must include an empty array at the end. For some reason, it didn't fail for me locally. Anyway I've already fixed it

packages/pretty-reporter/src/collapse-diagnostics.test.ts
daniilsapa: But it does change the order within one rule because I sort them to put errors first. I decided it was the quickest and most graceful way to add severity to consideration when we collapse diagnostics. In this way, we take errors first.

daniilsapa: As for the function that generates dummy diagnostics, it's a good idea, I've followed it.

packages/pretty-reporter/src/collapse-diagnostics.ts
```diff
@@ -0,0 +1,51 @@
+import { Diagnostic } from '@steiger/types'
+
+const DIAGNOSTIC_QUOTA = 20
+
+function distributeQuota(buckets: Array<number>, quota: number) {
+  const allItems = buckets.reduce((acc, bucket) => acc + bucket, 0)
+  const quotaPerBucket = buckets.slice(0).fill(0)
+  let remainingQuota = quota
+
+  for (let i = 0; remainingQuota > 0; i += 1) {
```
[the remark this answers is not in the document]
daniilsapa: Yes, I agree

daniilsapa:
It looks like I fixed all the issues, please review when you have time

packages/pretty-reporter/src/collapse-diagnostics.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (numOfItemsInBucket === undefined) {
+      i = 0
+      continue
+    }
+
+    // If the bucket already has the quota distributed for it or does not contain any items, skip it
+    if (assignedQuotaForBucket < numOfItemsInBucket && numOfItemsInBucket !== 0) {
+      quotaPerBucket[i] += 1
+      remainingQuota -= 1
+    }
```
illright, line 25: suggestion (non-blocking): I believe we can remove the `numOfItemsInBucket !== 0` part because `assignedQuotaForBucket` is always non-negative, so if `numOfItemsInBucket == 0`, then the first part of the condition will fail too (non-negative numbers are never less than zero).

I also suggest removing the comment for two reasons:
- with just one condition in the `if`, I think the code is pretty self-explanatory
- the comment describes the opposite of what's happening in the code below, which is a bit counter-intuitive
```suggestion
    if (assignedQuotaForBucket < numOfItemsInBucket) {
      quotaPerBucket[i] += 1
      remainingQuota -= 1
    }
```

illright (changes requested):
Yep, now the code is good! There is one potential issue in the workflow, and I figured I'd also throw in a little non-blocking suggestion as well

packages/pretty-reporter/src/collapse-diagnostics.ts
daniilsapa: I agree, you're right

illright (approved):
Okay, this seems good to go!

PR #168, opened 2025-01-13
Switch from `minimatch` to `micromatch`

### LINKED ISSUE #167: Replace `minimatch` with `micromatch`
Someone complained in the Telegram chat that they couldn't install Steiger with Yarn running Node 18 because `minimatch`, a dependency, has dropped Node 18 compatibility.

To solve this problem, we can switch to [`micromatch`](https://github.com/micromatch/micromatch), which:
- is more compatible
- weighs less
- is faster
- is a drop-in replacement, no extra thought needed

illright:
Closes #167


packages/steiger/src/shared/globs/create-filter-according-to-globs.ts
```diff
@@ -1,5 +1,5 @@
 import { isNegatedGlob } from './utilities'
-import { minimatch } from 'minimatch'
+import micromatch from 'micromatch'
 
 // ! Don't use platform specific path separators in the glob patterns for globby/minimatch
```
daniilsapa, line 4: Let's also replace the name of the library in the comment?

daniilsapa (approved):
Ha, I appreciated the title, honestly, it looks pretty funny and gave me a good laugh 😆. I wish we had some kind of "macromatch" or just "match"  earlier to make a whole sequence.

As for your solution, well done and it's the fastest possible!

PR #171, opened 2025-01-18
Upgrade to pnpm 10

illright:
Also upgrades other dependencies.

Here's the changelog for pnpm 10: https://github.com/pnpm/pnpm/releases/tag/v10.0.0

The main change is that `postinstall` scripts are no longer executed by default for security. Now we have to authorize each package that wants to run a build script. In our case, it's just `esbuild`.

Also pnpm 10 is better at versioning itself, which means that if you have pnpm 10 on your system, but pnpm 9 in some project, everything will just work seamlessly.


daniilsapa (approved):
Great! Thanks

daniilsapa (approved):

PR #173, opened 2025-02-02
Separate Vitest-dependent utils into a different entrypoint

illright:
`@steiger/toolkit` has some utilities to help plugin writers write tests easier. We don't want to force people to use Vitest, so it's marked as an optional peer dependency. However, the file you import when you import `@steiger/toolkit` contains an import from `vitest`, which fails when you don't have Vitest installed.

This PR creates another entrypoint, `@steiger/toolkit/test`. Now this entrypoint has all the test utils, and you can only import it if you have Vitest installed. If you don't need test utils, you can import everything else from `@steiger/toolkit`, no Vitest required.

### DIFF
--- a/packages/toolkit/src/index.ts
+++ b/packages/toolkit/src/index.ts
@@ -4,5 +4,3 @@ export { findAllRecursively } from './find-all-recursively.js'
 export { enableAllRules, createConfigs } from './create-configs.js'
 export { createPlugin } from './create-plugin.js'
 export type { ConfigObjectOf } from './config-object-of.js'
-
-export { compareMessages, createFsMocks, joinFromRoot, parseIntoFolder } from './prepare-test.js'
--- a/.changeset/curly-years-sparkle.md
+++ b/.changeset/curly-years-sparkle.md
@@ -1,5 +1,5 @@
 ---
-'@steiger/toolkit': patch
+'@steiger/toolkit': minor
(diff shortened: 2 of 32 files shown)


packages/toolkit/src/index.ts
```diff
@@ -4,5 +4,3 @@ export { findAllRecursively } from './find-all-recursively.js'
 export { enableAllRules, createConfigs } from './create-configs.js'
```
daniilsapa, line 4: `findAllRecursively` still has some references to vitest (`import.meta.vitest`) and is still imported in `@steiger/toolkit`. But looks like it shouldn't cause any problems because `import.meta.vitest` will be "undefined"

daniilsapa (approved):
Yeah, that's a pretty subtle moment but makes perfect sense. Thanks! 

I don't have any questions except one. Should we cover the fact that `@steiger/toolkit/test` requires Vitest to be installed (maybe a short note somewhere in README). Not a big deal, we can put this aside until we have full-fledged documentation for plugin creators.

packages/toolkit/src/index.ts
illright: Yeah, I think it should be fine

illright:
The point about docs is good, but I agree, it can wait until we have some kind of docs :D

PR #181, opened 2025-03-19
Support for Multiple Index Files

### LINKED ISSUE #180: SSR problem (Remix)
Fullstack frameworks have server-side and client-side components/features, and the server-side ones should not overlap with the client-side ones. Public API does not imply such a separation and the rule does not allow you to specify an additional Public API file, otherwise you will have to disable this rule on almost all layers

Image

In summary, it is desirable to support multiple entry files for slice, with a name in the steiger config

gwbaik9717:
### Description
This PR allows support for multiple index files. 

Also, index files other than `index.js` are supported such as `index.client.js` and `index.server.js`.

To enable these improvements, the `@steiger/filesystem` package has been upgraded to `v3.0.0`. As part of this update, the `getIndex` function has been replaced with the new `getIndexes` function.

### Fixes
Fixes #180

### DIFF
--- a/.changeset/grumpy-snails-relax.md
+++ b/.changeset/grumpy-snails-relax.md
@@ -1,0 +1,5 @@
+---
+'@feature-sliced/steiger-plugin': patch
+---
+
+Added support for multiple public APIs and enhanced logic for identifying public APIs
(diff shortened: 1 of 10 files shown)


illright (changes requested):
Thanks for the PR! Sorry it took so long to get a review, I'm a bit less active lately. Everything looks good, but I noticed that the `no-layer-public-api` rule does not have a test for the added code, could you add that please?

illright (approved):

illright:
Thank you! Let's get this merged

PR #188, opened 2025-04-06
Fix Bundle-Size Workflow to correctly checkout forked branch

### LINKED ISSUE #184: The bundle size workflow doesn't work with forks
See #181 for an example

gwbaik9717:
### Description
This PR fixes the workflow of `Bundle size` by ensuring that both the `base` branch and the `PR branch` **are correctly fetched**.

image

Previously, both branches were fetched from the **same repository**, which caused issues when the PR came from a fork. This update fixes the logic to dynamically determine the correct ref (branch) and repository for each case, allowing us to:
	•	Check out the base branch from the upstream repository
	•	Check out the PR branch from the contributor’s fork

### Fixes
Fixes #184

### DIFF
--- a/.github/workflows/bundle-size-trusted.yaml
+++ b/.github/workflows/bundle-size-trusted.yaml
@@ -62,9 +62,23 @@ jobs:
             core.summary.write();
             return report;
 
+      - name: Extract PR number
+        id: extract-pr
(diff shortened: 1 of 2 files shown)


gwbaik9717:
Still figuring out `Bundle Size (trusted)` workflow.

PR #191, opened 2025-04-19
handle exception when dependencies are not installed

### LINKED ISSUE #91: Uncaught exception when dependencies are not installed
```
file://pnpm/global/5/.pnpm/tsconfck@3.1.1_typescript@5.5.3/node_modules/tsconfck/src/parse.js:261
        throw new TSConfckParseError(
              ^

TSConfckParseError: failed to resolve "extends":"./.nuxt/tsconfig.json" in 
    code: 'MODULE_NOT_FOUND',
    requireStack: [ '' ]
  },
  tsconfigFile: ''
}
```

blasdfaa:
resolves #91


blasdfaa:
@illright could you please review?

packages/steiger/src/features/handle-error.ts
```diff
@@ -0,0 +1,48 @@
...
+    const packageJson = readPackageJSON(projectRootPath)
+    const hasNuxt = !!(packageJson.dependencies?.nuxt || packageJson.devDependencies?.nuxt)
+    if (!hasNuxt) return false
+
+    console.error('\n\x1b[31mError: Unable to find Nuxt TypeScript configuration\x1b[0m')
+    console.error('\nThis appears to be a Nuxt project, but the TypeScript configuration files are missing.')
+    console.error('These files are auto-generated when you install dependencies.')
+    console.error('\nTo fix this:')
+    console.error('1. Run \x1b[36mnpm install\x1b[0m or \x1b[36myarn\x1b[0m or \x1b[36mpnpm install\x1b[0m')
+    console.error('2. Then run Steiger again\n')
```
illright, line 33: suggestion: we have `picocolors` installed, let's use that instead of raw ANSI codes for text coloring for better readability

packages/steiger/src/features/handle-error.ts
```diff
@@ -0,0 +1,48 @@
...
+interface ErrorContext {
+  vfs: Folder
+}
+
+interface PackageJson {
+  dependencies?: Record<string, string>
+  devDependencies?: Record<string, string>
+}
+
+const readPackageJSON = (path: string) => JSON.parse(readFileSync(join(path, 'package.json'), 'utf-8')) as PackageJson
```
illright, line 15: issue: it might be the case that `package.json` is in one of the parent folders, not necessarily the one where Steiger was run

suggestion: let's use `empathic` to locate the `package.json` file:

```
import * as find from 'empathic/find'
find.up('package.json', { cwd: path })
```

packages/steiger/src/features/handle-error.ts
```diff
@@ -0,0 +1,48 @@
...
+    console.error('1. Run \x1b[36mnpm install\x1b[0m or \x1b[36myarn\x1b[0m or \x1b[36mpnpm install\x1b[0m')
+    console.error('2. Then run Steiger again\n')
+    return true
+  } catch {
+    return false
+  }
+}
+
+export function handleError(error: unknown, context: ErrorContext): never {
+  const isHandled = [checkNuxtConfigError].some((checker) => checker(error, context))
```
illright, line 41: praise: interesting architecture! I'm not sure how many more of those there will be, but thanks for building with extensibility in mind :)

packages/steiger/src/features/handle-error.ts
```diff
@@ -0,0 +1,48 @@
...
+    console.error('\nTo fix this:')
+    console.error('1. Run \x1b[36mnpm install\x1b[0m or \x1b[36myarn\x1b[0m or \x1b[36mpnpm install\x1b[0m')
+    console.error('2. Then run Steiger again\n')
+    return true
+  } catch {
+    return false
+  }
+}
+
+export function handleError(error: unknown, context: ErrorContext): never {
```
illright, line 40: issue: it takes a bit of code reading to understand what this function does, especially since its return type is `never`

suggestion: let's add a docstring that explains that this function attempts to gracefully handle the error using one of the known error checkers, and rethrows it if it can't

packages/steiger/src/features/handle-error.ts
```diff
@@ -0,0 +1,48 @@
...
+  const isHandled = [checkNuxtConfigError].some((checker) => checker(error, context))
+
+  if (isHandled) {
+    process.exit(1)
```
illright, line 44: issue: not a big fan of handling process exits from within the code of the linter app. This is more the responsibility of the CLI — to print the error message and exit with an error code

suggestion: maybe we can redo this function to instead throw a known error, and also export a separate function that would handle the error message printing? The way I see it working is like this:
1. The linter code encounters a tsconfig error
2. The `handleError` function recognizes that this is a Nuxt-related issue
3. The linter code throws a `KnownIssueError` (or something) to the caller, in this case, `cli.ts`
4. The CLI uses the the code from the `handle-error` feature to print some guidance
5. The CLI returns with a unique error code

In this solution, I think it would also make sense to make `handle-error` a folder, and maybe also give it a more descriptive name like `detect-known-exceptions`

examples/nuxt-app/public/favicon.ico
illright, line 1: suggestion: let's remove this and some other unnecessary template files: `README.md`, `robots.txt`

illright (changes requested):
Thanks for tackling this issue! I like the overall direction of the solution, left some comments to align it more to how the current components (linter, CLI, features) interact.

Could you also add a patch changeset for this feature?

PR #201, opened 2025-06-29
add processConfiguration to export

ArtMan-8:
Resolves #200 

Added [processConfiguration](https://github.com/feature-sliced/steiger/blob/f801cb9b00ee3ae6ecda9a517b8fe55be149dfc6/packages/steiger/src/models/config/index.ts#L34) to exports. This makes it possible to set the configuration from anywhere.


ArtMan-8:
@illright could you please review?

.changeset/fifty-teachers-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'steiger': minor
```
illright, line 2: suggestion: let's make this a patch change. For one, it technically fixes something that should've been working all along, for two, I have larger plans for the next minor version, so I'd prefer to push it out as a patch change
```suggestion
'steiger': patch
```

illright (changes requested):
Yep, that should probably make the linter at least usable without the CLI :D

illright:
Released in 0.5.8! Thanks!

PR #205, opened 2025-07-20
chore: replace cosmiconfig with lilconfig

yslpn:
Replaced cosmiconfig with lilconfig to reduce the number of dependencies.

https://node-modules.dev/#install=cosmiconfig
vs
https://node-modules.dev/#install=lilconfig

846kb vs 17kb


illright (changes requested):
Great change, thanks! Could you please run `pnpm exec manypkg fix` to clear the CI error?

PR #208, opened 2025-07-28
chore: replace micromatch with picomatch

yslpn:
Improve glob filtering performance by replacing micromatch with picomatch and decrease node_modules size 
 
 234kb vs 83kb
 
 https://node-modules.dev/grid/depth#install=picomatch
 https://node-modules.dev/grid/depth#install=micromatch

### DIFF
--- a/packages/steiger/src/shared/globs/create-filter-according-to-globs.spec.ts
+++ b/packages/steiger/src/shared/globs/create-filter-according-to-globs.spec.ts
@@ -214,6 +214,40 @@ describe('createFilterAccordingToGlobs', () => {
     expect(filteredFiles).toEqual(expected)
   })
 
+  it('should correctly handle brace expansion in inclusion patterns', () => {
+    const files = ['/src/shared/lib/feature-a.ts', '/src/shared/lib/feature-a.tsx', '/src/shared/lib/feature-a.js']
+
+    const expected = ['/src/shared/lib/feature-a.ts', '/src/shared/lib/feature-a.tsx']
+
+    const filter = createFilterAccordingToGlobs({
+      inclusions: ['**/*.{ts,tsx}'],
+    })
+
+    const filteredFiles = files.filter(filter)
+
+    expect(filteredFiles).toEqual(expected)
+  })
+
+  it('should correctly handle brace expansion in negated exclusion patterns', () => {
+    const files = [
+      '/src/shared/lib/keep.ts',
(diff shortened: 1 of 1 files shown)


packages/steiger/src/shared/globs/create-filter-according-to-globs.ts
```diff
@@ -13,29 +13,29 @@ export function createFilterAccordingToGlobs({ inclusions, exclusions }: ApplyGl
...
 
-  function filterAccordingToGlobs(path: string) {
-    const matchesInclusionPatterns =
-      !thereAreInclusions || inclusions.some((pattern) => micromatch.isMatch(path, pattern))
-    let isIgnored = false
+  const isIncluded = thereAreInclusions ? picomatch(inclusions, picomatchOptions) : () => true
+
+  const positiveExclusionPatterns =
+    (thereAreExclusions && exclusions.filter((pattern) => !isNegatedGlob(pattern))) || []
+  const negativeExclusionPatterns =
```
illright, line 22: suggestion (non-blocking): since "positive exclusion" and "negative exclusion" aren't commonly understood terms, hence they require some time to understand, perhaps we could instead have "positive exclusion" and "re-inclusion"?

```suggestion
  const reInclusionPatterns =
```

illright (dismissed):
Code-wise I think it's great!

packages/steiger/src/shared/globs/create-filter-according-to-globs.ts
yslpn: Agree. Fixed

illright (changes requested):
Thanks! The regression in the lilconfig migration made me a bit worried, could you add some extra tests to `packages/steiger/src/shared/globs/create-filter-according-to-globs.spec.ts` relating to brace expansion? Currently there's only one, and it only tests brace expansion in exclusions, I'd like to make sure it also works in inclusions and negated patterns.

yslpn:
> Thanks! The regression in the lilconfig migration made me a bit worried, could you add some extra tests to `packages/steiger/src/shared/globs/create-filter-according-to-globs.spec.ts` relating to brace expansion? Currently there's only one, and it only tests brace expansion in exclusions, I'd like to make sure it also works in inclusions and negated patterns.

Ok, thanks for the review. I'll add more tests later this week.

illright (approved):
Amazing, thanks a lot!

PR #215, opened 2025-08-12
feat(steiger-plugin): add no-wildcard-exports rule and fix tsconfig resolution

Gaic4o:
# Background 

- Implement `no-wildcard-exports` rule in `@feature-sliced/steiger-plugin`
- Disallow `export *` in public APIs of business layers
- Allow `export * as namespace` and wildcard exports in `shared/app`
- Public API is detected via `index.(js|jsx|ts|tsx);` non-index and test files are ignored
- Add tests and `README.md` for the rule
- Move `@tsconfig/node18` to `dependencies` in `@steiger/tsconfig` for proper tsconfig extends resolution

### DIFF
--- a/packages/steiger-plugin-fsd/package.json
+++ b/packages/steiger-plugin-fsd/package.json
@@ -43,6 +43,7 @@
     "@feature-sliced/filesystem": "^3.0.1",
     "fastest-levenshtein": "^1.0.16",
     "lodash-es": "^4.17.21",
+    "oxc-parser": "^0.47.1",
     "pluralize": "^8.0.0",
     "precinct": "^12.2.0",
     "tsconfck": "^3.1.6"
--- a/packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md
+++ b/packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md
@@ -1,0 +1,61 @@
+# `no-wildcard-exports`
+
+Forbid wildcard exports (`export *`) in public APIs of business logic layers. Named exports and namespace exports (`export * as namespace`) are allowed.
+
+**Exception:** Wildcard exports are allowed in unsliced layers (`shared` and `app`), as they serve as foundational layers with different architectural purposes.
+
+This rule treats files named `index.js`, `index.jsx`, `index.ts`, `index.tsx` as the public API of a folder (slice/segment root). Non-index files (including test files like `*.spec.ts`, `*.test.ts`) are ignored.
+
+Examples of exports that pass this rule:
+
+```ts
+// Named exports (business logic layers)
+export { Button } from './Button'
+export { UserCard, UserAvatar } from './components'
+
+// Namespace exports (all layers)
+export * as userModel from './model'
+export * as positions from './tooltip-positions'
+
+// Wildcard exports (unsliced layers: shared, app)
+// shared/ui/index.ts
+export * from './Button'
+export * from './Modal'
+export * from './Input'
+
+// shared/api/index.ts
+export * from './endpoints/auth'
+export * from './endpoints/users'
+
+// app/providers/index.ts
+export * from './AuthProvider'
+export * from './ThemeProvider'
+export * from './RouterProvider'
+```
+
+Examples of exports that fail this rule:
+
+```ts
+// ❌ Wildcard exports in business logic layers
+// entities/user/index.ts
+export * from './model'
+export * from './ui'
+
+// features/auth/index.ts
+export * from './ui'
+export * from './api'
+```
+
+## Rationale
+
+Wildcard exports in business logic layers make it harder to track which exact entities are being exported from a module. This can lead to:
+
+- Unintentionally exposing internal implementation details
+- Difficulty in tracking dependencies between modules
+- Potential naming conflicts when multiple modules use wildcard exports
+
+Using named exports or namespace exports makes the public API more explicit and easier to maintain in business logic layers.
+
+## Autofix
+
+This rule provides a suggested fix for wildcard exports in public API files by replacing them with a named export template (e.g. `export { ComponentA, ComponentB } from './components'`).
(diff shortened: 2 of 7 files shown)


packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md
```diff
@@ -0,0 +1,61 @@
+# `no-wildcard-exports`
+
+Forbid wildcard exports (`export *`) in public APIs of business logic layers. Named exports and namespace exports (`export * as namespace`) are allowed.
```
illright, line 3: issue: I'm a bit confused by the "business logic layers" here — I would consider `entities` and `features` to be more business-logic-layers than, say, `pages` and `widgets`

suggestion: perhaps you meant "all layers except `shared` and `app`"? Then I think _unsliced layers_ would be a better term, and to make it extra clear, we can explain what it means:
```suggestion
Forbid wildcard exports (`export *`) in public APIs of unsliced layers (i. e. all layers except Shared and App). Named exports and namespace exports (`export * as namespace`) are allowed.
```

packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+**Exception:** Wildcard exports are allowed in unsliced layers (`shared` and `app`), as they serve as foundational layers with different architectural purposes.
```
illright, line 5: issue: I would actually argue against allowing wildcard exports even in Shared and App. To me, wildcard exports conceal the public API of a group of modules, which really hurts the ability to discover a codebase shallowly, i. e. without digging into every file. 

question: why do you think they should be allowed there? And do you agree with the content of the [Public API reference page in the FSD docs](https://feature-sliced.github.io/documentation/docs/reference/public-api)?

packages/steiger-plugin-fsd/src/no-wildcard-exports/index.ts
```diff
@@ -0,0 +1,65 @@
...
+      const file = sourceFile.file
+
+      if (!/\.(js|jsx|ts|tsx)$/.test(file.path)) continue
+
+      // Allow wildcard exports in unsliced layers (shared, app)
+      if (sourceFile.layerName === 'shared' || sourceFile.layerName === 'app') continue
+
+      // Check if this is a public API file (typically index.* files)
+      const fileName = basename(file.path)
+      const isPublicApiFile = /^index\.(js|jsx|ts|tsx)$/.test(fileName)
```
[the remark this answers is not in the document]
illright: This is also important because we allow `index.client.ts` and `index.server.ts` as valid Public APIs, and this function takes that into account

packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md
```diff
@@ -0,0 +1,61 @@
...
+export * from './api'
+```
+
+## Rationale
+
+Wildcard exports in business logic layers make it harder to track which exact entities are being exported from a module. This can lead to:
+
+- Unintentionally exposing internal implementation details
+- Difficulty in tracking dependencies between modules
+- Potential naming conflicts when multiple modules use wildcard exports
```
illright, line 55: suggestion: I would also add the point about hurting shallow discovery (if you agree with it, that is :D)

packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Using named exports or namespace exports makes the public API more explicit and easier to maintain in business logic layers.
```
illright, line 57: suggestion: let's also leave a link to the [docs about Public API](https://feature-sliced.github.io/documentation/docs/reference/public-api)

packages/steiger-plugin-fsd/package.json
```diff
@@ -43,6 +43,7 @@
     "@feature-sliced/filesystem": "^3.0.1",
     "fastest-levenshtein": "^1.0.16",
     "lodash-es": "^4.17.21",
+    "oxc-parser": "^0.47.1",
```
illright, line 46: praise: kudos for using the oxc-parser :) I was hoping that we would switch to it entirely at some point and drop `precinct` because it isn't flexible enough and doesn't extend well

packages/steiger-plugin-fsd/src/no-wildcard-exports/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      // Skip files that are not public API
+      if (!isPublicApiFile) continue
+
+      // Parse file content using oxc-parser
+      const content = (file as FileWithContent).content
```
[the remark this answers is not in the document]
illright: I just realized — we don't include `content` as a field on the VFS node, we might need to explicitly read the file contents with `readFile`

packages/steiger-plugin-fsd/src/no-wildcard-exports/index.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+        } else if (child.path === joinFromRoot('entities', 'user', 'index.ts')) {
+          fileWithContent.content = "export * as userModel from './model'"
+        } else {
+          fileWithContent.content = ''
+        }
+      } else if (child.type === 'folder') {
+        addContentToFiles(child)
+      }
+    }
+  }
```
illright, line 130: issue: I think this conceals the fact that this test will actually not pass during real usage, since the linter engine doesn't include the `content` field

suggestion: let's use FS mocking for tests, like it's done in the [test file for `forbidden-imports`](https://github.com/feature-sliced/steiger/blob/1281d600506ae5f58479ca1266c53201c3a44247/packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts)

illright (changes requested):
Thanks for the great work! You covered nearly all aspects of creating a rule all by yourself :) The last things to do are to `packages/steiger-plugin-fsd/src/index.ts` and  add it to `disabledRules` (we don't want to enable it by default for now not to break people's CI, we will enable it by default in 0.6, with the next round of breaking changes) and also add it to the table of rules in the README (note that the README is copy-pasted between the project root and `packages/steiger`, so both files need to be updated).

PR #222, opened 2026-02-12
fix: use fallback characters for std output snapshots

Solant:
Chars from `figures` package depend on the current Windows shell, so snapshots will differ between different environments (currently tests were OK in cmd, but were failing in PowerShell)

I removed all symbols that could theoretically cause this issue

### DIFF
--- a/integration-tests/tests/__snapshots__/smoke-stderr-posix.txt
+++ b/integration-tests/tests/__snapshots__/smoke-stderr-posix.txt
@@ -1,42 +1,42 @@
 
 ┌ src/entities/user/api/getUser.ts
- Forbidden import from higher layer "app".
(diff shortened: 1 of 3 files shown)


illright:
Is that the reason why the tests are failing in that other PR? I figured that the environment would be consistent in CI

If the environment is indeed inconsistent, I would rather that we address that

Solant:
I don't think it is really feasible, because the issue is that the tick symbol for Windows in PowerShell and CMD is different. `figures` package is doing a great job of rendering the correct tick in the respective shell, but mock files can't be used in this case, because if the mock file and test will be run in different Windows shells, they won't match

illright (approved):

Gaic4o (approved):

PR #197, opened 2025-05-11
Discover plugins automatically

illright:
This PR marks an important internal architectural milestone — decoupling the `steiger` package from the `@feature-sliced/steiger-plugin` dependency while still retaining zero-config operation.

Help wanted to diagnose and fix the test failures on Windows

PR #228, opened 2026-04-18
fix(plugin): add root export entry for steiger plugin

### LINKED ISSUE #225: Error [ERR_PACKAGE_PATH_NOT_EXPORTED] for @feature-sliced/steiger-plugin
Hello, I want to intergrate steiger in my project on Vue, but I faced with the problem in steiger.config.ts. I am following the instructions on the official webside [npm](https://www.npmjs.com/package/steiger), but the problem remains.

Image

`Resolve error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/khavronin/Downloads/GoalsTracker/vueGoalsTracker/node_modules/@feature-sliced/steiger-plugin/package.json`

Node version - v22.14.0
Packages:
`"steiger": "^0.5.11"`
`"@feature-sliced/steiger-plugin": "^0.5.7",`

Probably the problem in the package @feature-sliced/steiger-plugin in node_modules
` "exports": {
      "types": "./dist/app.d.ts",
      "import": "./dist/app.mjs"
  },`

Instead of
`"exports": {
    ".": {
      "types": "./dist/app.d.ts",
      "import": "./dist/app.mjs"
    }
  },`

yslpn:
Fix package root exports for `@feature-sliced/steiger-plugin`.

Closes #225

### DIFF
--- a/.changeset/great-numbers-listen.md
+++ b/.changeset/great-numbers-listen.md
@@ -1,0 +1,5 @@
+---
+'@feature-sliced/steiger-plugin': patch
+---
+
+Fix package root exports in package.json
--- a/packages/steiger-plugin-fsd/package.json
+++ b/packages/steiger-plugin-fsd/package.json
@@ -12,8 +12,10 @@
     "typecheck": "tsc --noEmit"
   },
   "exports": {
-    "types": "./dist/index.d.ts",
-    "import": "./dist/index.js"
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    }
   },
   "files": [
     "dist",


Solant (approved):

Solant (commented):
The current conditional export is valid, as per https://nodejs.org/api/packages.html#conditional-exports

It also properly resolves locally, so I don't think we need this change.

illright:
Does this change actually solve the problem in that issue, btw? Didn't try myself yet, just wondering

Solant:
@illright I checked it myself; the problem in the referenced issue is due to pnpm hoisting, not conditional export.

The current conditional export works correctly.

illright:
Yeah, but pnpm is the industry standard now, and package isolation is one of its central selling points, if it doesn't work with pnpm out of the box, then it's an issue on our side

Solant:
In this case we need to remove `@feature-sliced/steiger-plugin` from the steiger `package.json` dependency and install it alongside `steiger` during initial setup, like `pnpm add -D steiger @feature-sliced/steiger-plugin`, and then create an initial config file like this:

```js
import { defineConfig } from 'steiger'

export default defineConfig([
  ...fsd.configs.recommended,
])
```

Because the current zero-config approach relies heavily on hoisting (which is disabled by default in pnpm). I think we can do some auto resolutions in `steiger` to omit the initial config requirement, though.

It looks like you tried to do something like that in https://github.com/feature-sliced/steiger/pull/197

illright:
> Because the current zero-config approach relies heavily on hoisting (which is disabled by default in pnpm).

If it's your direct dependency, then pnpm will put it on the top level of `node_modules`, so the zero-config approach is still perfectly compatible with pnpm. It will break if, say, someone only installs `steiger` and not `@feature-sliced/steiger-plugin`

Solant:
Ah, yes, we have a `@feature-sliced/steiger-plugin` as part of the installation process. All good then :smile: 

But I still don't think we need this `export` change

yslpn:
> The current conditional export is valid, as per https://nodejs.org/api/packages.html#conditional-exports
> 
> It also properly resolves locally, so I don't think we need this change.

@Solant Thanks for review. I think you're right. The reason I decided to make this change is because I looked at all the exports for the monorepo, and they all use

```json
  "exports": {
    ".": {
```

I wanted to make it consistent and fix the problem of cource.

In any case, we can close the issue and PR.

illright:
Let's wait with closing until we hear from the issue author :)

PR #235, opened 2026-05-14
feat: replace precinct with tree-sitter

### LINKED ISSUE #116: Support Svelte
We're using the [`precinct`](https://github.com/dependents/node-precinct) package to extract dependencies, and it currently doesn't support Svelte files. Figure out a way to add support for Svelte, either by contributing to the precinct package, or by replacing it for something that does support Svelte

Solant:
Replace the `precinct` library for dependency extraction with `tree-sitter`. This adds support for `svelte` and `astro` files, implements more resilient parsing logic (no errors are thrown on incomplete/invalid source files), and reduces install size.

The current implementation uses the WASM version of parsers for compatibility reasons, though it is not the fastest for Node runtime.


Gaic4o (dismissed):

packages/steiger-plugin-fsd/src/_language-tools/astro.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { expect, it } from 'vitest'
+import { extractDependencies } from './index.js'
+
+it('extracts esm dependencies from Svelte source code', async () => {
```
illright: issue (non-blocking): typo
```suggestion
it('extracts esm dependencies from Astro source code', async () => {
```

packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
```diff
@@ -28,7 +28,21 @@ vi.mock('node:fs', async (importOriginal) => {
   return createFsMocks(
     {
       '/src/shared/ui/styles.ts': '',
-      '/src/shared/ui/Button.tsx': 'import styles from "./styles";',
+      // '/src/shared/ui/Button.tsx': 'import styles from "./styles";',
```
illright: question: did you intend to leave this change in?

illright (dismissed):
Surprisingly beautiful! I expected a lot more faff with tree-sitter :D

packages/steiger-plugin-fsd/src/_language-tools/astro.spec.ts
Solant: Nice catch, fixed

illright (approved):

packages/steiger-plugin-fsd/src/forbidden-imports/index.spec.ts
Solant: This is a WIP leftover; I reverted the mock back to the original value

Gaic4o (approved):

PR #236, opened 2026-05-24
build: cleanup package.json for published packages

Solant:


Gaic4o (approved):

packages/steiger/package.json
```diff
@@ -42,8 +43,7 @@
   },
   "files": [
     "dist",
-    "src",
-    "README.md"
```
illright, line 46: question: why did you decide to remove README from the published files?

packages/steiger/LICENSE
illright, line 1: nit (non-blocking): I think the convention for license files is to not have any extension, but I don't think it matters too much.

Btw, should this file also be included in the `files` of the package.json?

package.json
```diff
@@ -22,6 +22,7 @@
     "@manypkg/cli": "^0.23.0",
     "@steiger/eslint-config": "workspace:*",
     "eslint": "^9.32.0",
+    "eslint-plugin-package-json": "^1.1.0",
```
illright, line 25: suggestion: there is another specialized project called publint, I used it in the past and was pretty happy with it. Not saying we should use it instead of ESLint here, but just FYI, in case you haven't heard of it

packages/steiger/package.json
Solant: This file was marked by linter https://eslint-plugin-package-json.dev/rules/no-redundant-files/

It looks like some files are included in the tarball automatically, so there is no need to add them

packages/steiger/LICENSE
Solant: I checked the resulting tarball, and `LICENSE.md` was included just like regular `LICENSE`. But you are right, I changed the file name to match the convention.

This file is also always included in the tarball no matter what https://eslint-plugin-package-json.dev/rules/no-redundant-files/

illright (approved):

package.json
Solant: Yeah, I haven't heard about this `publint` before. It looks like they are very similar; it seems like the ESLint plugin is a superset of `publint`

For example, `authors` is not a valid key, according to the npm docs https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#people-fields-author-contributors and `publint` doesn't lint this as an issue https://publint.dev/steiger@0.5.12

But this is up to you; I'm fine with any tool. The ESLint plugin was just very easy to add without any additional integrations, and it covered the majority of cases

illright: I'm cool with ESLint, let's keep it

PR #241, opened 2026-06-29
feat: add performance profiling

Solant:
Add performance profiling

### DIFF
--- a/.changeset/chatty-teeth-bake.md
+++ b/.changeset/chatty-teeth-bake.md
@@ -1,0 +1,5 @@
+---
+'steiger': minor
+---
+
+Add performance profiling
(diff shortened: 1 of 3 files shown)


Gaic4o (approved):

illright (approved):
Ideally, we would not even measure time if timing is disabled since it does incur a small performance overhead, but it's not a big deal

Solant:
> Ideally, we would not even measure time if timing is disabled since it does incur a small performance overhead, but it's not a big deal

Agree, this is a very barebones implementation. This is just to verify the performance impact of a couple of CPU-intensive rules for fs caching that I'm currently working on. It also might be beneficial for other developers to identify slowdowns.

PR #249, opened 2026-07-16
feat: add language server implementation

Solant:


packages/types/index.ts
illright, line 1: issue: this looks kinda weird, to be honest. I get that it is to prevent breaking changes, but we are not 1.0.0 yet, so I wouldn't worry too much about it in favor of a well-designed API for 1.0.0

suggestion: let's define start as a separate field? Or alternatively, to avoid breaking changes, make `line?: number | { start: number; end: number }`, and the same for column?

illright (commented):
wow, I am amazed at how little changes were required, great job! One comment about the API

packages/types/index.ts
Solant: done

Gaic4o (approved):

illright (approved):

PR #252, opened 2026-08-25
feat: add the no-wildcard-exports rule

Gaic4o:
# Background 

Added a `no-wildcard-exports` rule that detects wildcard re-exports in public APIs.

It reports the following patterns:

```ts
export * from "./foo";
export type * from "./foo";
export * as foo from "./foo";
export type * as foo from "./foo";
```

Explicit re-exports are still allowed:

```ts
export { foo } from "./foo";
export { foo as bar } from "./foo";
```

To support this, `_language-tools` was refactored to analyze imports and re-exports together and reuse the same FS cache. The behavior of existing import-based rules remains unchanged.

### DIFF
--- a/packages/steiger-plugin-fsd/src/_language-tools/index.ts
+++ b/packages/steiger-plugin-fsd/src/_language-tools/index.ts
@@ -2,7 +2,7 @@ import { join, extname, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { isBuiltin } from 'node:module'
 import { readFileSync } from 'node:fs'
-import { Parser, Query, Language, Range, type Tree } from 'web-tree-sitter'
+import { Parser, Query, Language, Range, type Node, type Tree } from 'web-tree-sitter'
 import { createFSCache } from '../_lib/fs-cache.js'
 
 // TODO: replace with import.meta.dirname when upgrading to nodejs 20/22
@@ -43,9 +43,10 @@ interface Extractor {
   extensions: string[]
   language: Language
   injections: Array<{ query: Query; lang: string }>
-  queries: Array<{ query: Query; type: 'static' | 'dynamic' }>
-  /** Matches `export * from` statements, capturing the whole statement as `@statement` and the module specifier as `@path`. */
-  wildcardExportQuery?: Query
+  /** Queries for the modules a file pulls in. `static`/`dynamic` describes how the module is loaded. */
+  importQueries: Array<{ query: Query; type: 'static' | 'dynamic' }>
+  /** Queries for the modules a file re-exports. Each one captures `@statement` and `@source`, and produces one `kind`. */
+  reExportQueries: Array<{ query: Query; kind: ReExportInfo['kind'] }>
 }
 
 const extractors: Array<Extractor> = [
@@ -53,9 +54,9 @@ const extractors: Array<Extractor> = [
     type: 'tsx',
     extensions: ['.tsx', '.jsx', '.ts', '.js', '.cjs', '.mjs'],
     language: tsx,
-    queries: [
+    importQueries: [
       {
-        query: new Query(tsx, '(import_statement source: (string (string_fragment) @path))'),
+        query: new Query(tsx, '(import_statement source: (string (string_fragment) @source))'),
         type: 'static',
       },
       {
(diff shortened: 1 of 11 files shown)


packages/steiger-plugin-fsd/src/_language-tools/index.ts
Solant, line 1: I like how we can add many new rules now using tree-sitter, but I don't think adding random flags and queries into `extractors` is the way to go. As this API was created specifically for getting _imports only_, and it doesn't really scale up to exports (especially with wildcards and namespaces). I propose we refactor this part a bit first to make sure that it handles both imports and exports simultaneously, and then add export-based rules on top of it. (this is also the case for #251)

1. make `extractDependencies` return something like 
```typescript
function extractDependencies(): Statement[]

// I'm not sure if the types are correct, need to veryfy how wildcards and namespaces work together
// this also doesn't take into account import renaming and import attributes
type Statement = Import | Export;
interface Import {
  specifier: string
  path: string;
}
interface Export {
  namespace?: string
  specifier: string
  path: string
}
```

2. save all import/export information back to the FS cache so we don't have to re-parse the file
3. add framework-specific logic for exports: for example, Vue files with the `setup` attribute by default have one implicit default export (I believe this is also the case for svelte)
4. (optionally) change the querying options for the `extractDependencies` so we don't have to add too much filtering logic into the rules

Solant (changes requested):

packages/steiger-plugin-fsd/src/_language-tools/index.ts
```diff
@@ -258,3 +311,21 @@ export async function extractDependencies(
     return true
   })
 }
+
+const wildcardExportCache = createFSCache<WildcardExport[]>()
+
+/**
+ * Find the wildcard re-exports (`export * from`) in a file.
+ *
+ * Namespace re-exports (`export * as ns from`) are not reported. They add one name to the module's
```
Solant: I think this could be a bit destructive; for example this file will be considered a valid public API file, while not providing any of contract/abstraction

```typescript
// index.ts
export * as ui from "./ui";
export * as model from "./model";
export * as api from "./api";
```

Gaic4o: `export * as foo from "./foo";`
`export type * as foo from "./foo";`

After hearing your feedback, I think it would be better to handle this part as well. Thank you for the review.

packages/steiger-plugin-fsd/src/_language-tools/index.ts
Gaic4o: I’ve updated the implementation based on your feedback, and namespace re-exports are now reported by this rule as well.

One thing I noticed while working on this is that issue #5 currently lists the following as an example that should pass:

`export * as positions from "./tooltip-positions";`

This seems to conflict with the direction of the current review. For now, I’ve followed the current review and included namespace re-exports in the rule, but could you confirm whether this is the intended behavior?

Gaic4o: I also looked into the framework specific implicit exports you mentioned. To implement this properly, I think I would need to spend some additional time looking into and verifying the behavior of Vue, Svelte.

Would you prefer to include this in the scope of this PR as well, or would it be better to keep this PR focused on the shared analysis structure and handle framework-specific implicit exports separately in a follow up?

Solant: Ok, thanks!

> This seems to conflict with the direction of the current review. For now, I’ve followed the current review and included namespace re-exports in the rule, but could you confirm whether this is the intended behavior?

My main concern is that exports like these on a public API level are a bit dangerous:

```typescript
export * as ui from "./ui"
export * as model from "./model"
```

So maybe we could potentially add a rule so that we don't re-export all segment content in this way?

> Would you prefer to include this in the scope of this PR as well, or would it be better to keep this PR focused on the shared analysis structure and handle framework-specific implicit exports separately in a follow up?

After second thought, I think we don't need to do the framework-specific export detection. As there are a ton of edge cases that we aren't really concerned about.

Gaic4o: Got it! I'll update the PR accordingly. Thanks for the review!

Solant (approved):

PR #262, opened 2026-09-17
fix: treat re-exports as dependencies

### LINKED ISSUE #261: Re-export statements are not collected as dependencies since @feature-sliced/steiger-plugin 0.6.0
### Summary

The rules that resolve module specifiers never see `export ... from` statements, so a re-export is not a dependency edge. This gives false negatives in `fsd/forbidden-imports` and `fsd/no-public-api-sidestep`, and a false positive in `fsd/insignificant-slice`. Up to `@feature-sliced/steiger-plugin` 0.5.8, shipped in `steiger@0.5.12`, these cases were reported.

### Reproduction

One minimal FSD fixture, relative specifiers only, run as `npx steiger ./src` on three installs: `steiger@0.5.12` with plugin 0.5.8 and precinct, `steiger@0.5.13` with plugin 0.6.0 and web-tree-sitter, `steiger@0.6.0` with plugin 0.7.0.

| case | 0.5.12 | 0.5.13 and 0.6.0 |
| --- | --- | --- |
| `entities/user/model/user.ts` with `export { login } from '../../../features/auth'` | forbidden import from a higher layer | no diagnostic |
| `features/a` with `export * from '../../b'` | forbidden cross-import from a sibling slice | no diagnostic |
| `features/policy` with `export { createUser } from '../../../entities/user/model/user'` | forbidden sidestep of the public API | no diagnostic |
| a slice referenced only through `export ... from` | no diagnostic | insignificant slice, reported as having no references |

Written as a plain `import`, every case is reported on all three versions, so the rules work and only the dependency edge is missing.

### Cause

`extractDependencies` in `packages/steiger-plugin-fsd/src/_language-tools/index.ts` runs tree-sitter queries for `import_statement`, `require` and dynamic `import()`. A `from` clause on an export is an `export_statement`, which no query matches. Probing the bundled tsx grammar directly on such a file shows no ERROR node, and the query `(export_statement source: (string (string_fragment) @path))` matches both `export { x } from` and `export * from`.

Before 0.6.0 the rules used `precinct.paperwork`, which returned export sources. The parser swap in #235 does not mention re-exports in its changeset, and there is no re-export case in `_language-tools/typescript.spec.ts`, so this looks unintentional.

### Expected

`export ... from`, `export * from`, `export * as ns from` and their type-only variants are collected as static dependencies.

### Environment

macOS, Node 24.19.0, npm 11.17.0, TypeScript sources. Vue, Svelte and Astro feed the same extractor, so they are probably affected as well, not verified.

conarti:
## Problem

`extractDependencies` collects only `import`, `require` and dynamic `import()`, so `export … from` statements are invisible to every rule that resolves module specifiers. This regressed in `@feature-sliced/steiger-plugin` 0.6.0, first shipped in `steiger@0.5.13`, when precinct, which returned export sources, was replaced.

Four cases, all reported by `steiger@0.5.12` and silent since:

- an upward re-export from `entities` to `features` is not reported
- a sibling slice re-export is not reported
- a public API sidestep written as a re-export is not reported
- a slice consumed only through re-exports is falsely reported by `fsd/insignificant-slice` as having no references

Closes #261.

## Change

A re-export names another module exactly like a static import does, so `extractDependencies` now returns both, merged in source order. Re-exports come back as static, non dynamic dependencies anchored at their specifier range, and `isBuiltin` is applied so `export { readFile } from 'node:fs'` is filtered out just like the import form.

`analyzeModule`, `extractReExports`, `ReExportInfo` and the tree-sitter queries are untouched, so `no-wildcard-exports` keeps reading the shape of a re-export as before. The docblock stating that re-exports are excluded is updated to match.

Based on #252, as suggested in the issue, so review the last commit only.

## Tests

- `_language-tools`: `export { a } from`, `export * from`, `export * as ns from` and the three type only variants, asserted as full dependency objects; a re-export preceding an import pins the source order; a built in re-export pins the filtering; the cache spec now expects imports and re-exports together; one Vue case covers the framework extractors.
- `forbidden-imports`: an entity re-exporting from a feature, and a slice re-exporting from a sibling slice.
- `no-public-api-sidestep`: a re-export reaching into another slice's internals.
- `insignificant-slice`: a slice referenced through re-exports from two slices is no longer reported.

Each of them fails without the change to `_language-tools/index.ts`, and removing either the sort or the built in filtering turns two of them red.

Verified end to end as well: built the packages and ran the binary over the reproduction fixtures from the issue. The four cases report again, a legal downward re-export through a public API stays clean, and so does the baseline project. The `kitchen-sink-of-fsd-issues` example is byte identical before and after.

## Two things for you to decide

- The changeset says `minor`, because projects will report errors they do not report today, and #235, which removed the behaviour, was also a minor. Happy to drop it to `patch` if you would rather ship it as a regression fix.
- With re-exports counted as references, a slice reached only through a single barrel moves from `This slice has no references` to `This slice has only one reference in slice "…"`. It is still reported, just differently. That follows from treating re-exports as ordinary imports, but say the word if you meant otherwise.

### DIFF
--- a/.changeset/no-wildcard-exports.md
+++ b/.changeset/no-wildcard-exports.md
@@ -2,4 +2,4 @@
 '@feature-sliced/steiger-plugin': minor
 ---
 
-add the `no-wildcard-exports` rule, which forbids `export * from` in public APIs but allows `export * as ns from`. It is disabled by default.
+add the `no-wildcard-exports` rule, which forbids `export * from` and `export * as ns from` in public APIs. The rule is disabled by default.
--- a/.changeset/no-wildcard-exports.md
+++ b/.changeset/no-wildcard-exports.md
@@ -1,0 +1,5 @@
+---
+'@feature-sliced/steiger-plugin': minor
+---
+
+add the `no-wildcard-exports` rule, which forbids `export * from` in public APIs but allows `export * as ns from`. It is disabled by default.
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ Currently, Steiger is not extendable with more rules, though that will change in
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/typo-in-layer-name/README.md"><code>fsd/typo-in-layer-name</code></a></td> <td>Ensure that all layers are named without any typos.</td> </tr>
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/no-processes/README.md"><code>fsd/no-processes</code></a></td> <td>Discourage the use of the deprecated Processes layer.</td> </tr>
   <tr> <td><a href="./packages/steiger-plugin-fsd/src/import-locality/README.md"><code>fsd/import-locality</code></a></td> <td>[disabled] Require that imports from the same slice be relative and imports from one slice to another be absolute.</td> </tr>
+  <tr> <td><a href="./packages/steiger-plugin-fsd/src/no-wildcard-exports/README.md"><code>fsd/no-wildcard-exports</code></a></td> <td>[disabled] Forbid wildcard re-exports (<code>export * from</code>) in public APIs.</td> </tr>
 </tbody>
 </table>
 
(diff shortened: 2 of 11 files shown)


Solant:
hey :wave: 

can you update your PR? I think the parsing part is already done in #252, so the final change should be relatively small

